Issue #42, opened 2025-08-14
[BUG] Installed Correctly but Unable to Fetch LinkedIn Profiles


marko-vasic:
## Installation Method
- [ x] Docker (specify docker image version/tag): _._._
- [ ] Claude Desktop DXT extension (specify docker image version/tag): _._._
- [ ] Local Python setup

## When does the error occur?
- [ ] At startup
- [x ] During tool call (specify which tool):
  - [x ] get_person_profile
  - [ ] get_company_profile
  - [ ] get_job_details
  - [ ] search_jobs
  - [ ] get_recommended_jobs
  - [ ] close_session

## MCP Client Configuration

**Claude Desktop Config** (`/Users/[username]/Library/Application Support/Claude/claude_desktop_config.json`):
```json
{
  "mcpServers": {
    "linkedin": {
      // Your configuration here (remove sensitive credentials)
    }
  }
}
```

## MCP Client Logs
**Claude Desktop Logs** (`/Users/[username]/Library/Logs/Claude/mcp-server-LinkedIn MCP Server.log`):
```
Paste relevant log entries here
```

## Error Description
What went wrong and what did you expect to happen?

I am getting an error that it is not possible to fetch account information of a linkedin profile.
It seems there are some permission issues.
Please take a look at the attachments and let me know if you have suggestions on how to resolve.

<img width="936" height="928" alt="Image" src="https://github.com/user-attachments/assets/7a5b8c7f-3f27-4d84-b68d-bf41f90a239e" />
<img width="1060" height="935" alt="Image" src="https://github.com/user-attachments/assets/e9b1bb08-eefe-4808-9891-0db890214662" />

marko-vasic:
I tried testing with mcp inspector,
and running get_person_profile tool returns this error
`"MCP error -32001: Request timed out"`

I don't see if there are additional logs.

stickerdaniel:
We'll transition to a new scraping library, this will resolve the missing field issue and speed up profile scraping. See #23 #43

Issue #23, opened 2025-07-12
[FEATURE] WIP: transition get_profile tool to fast-linkedin-scraper


stickerdaniel:
## Feature Description
Refactor the get profile tool to use [fast-linkedIn-scraper](https://github.com/stickerdaniel/fast-linkedin-scraper) instead of [linkedin_scraper](https://github.com/joeyism/linkedin_scraper) for improved performance and scraping results.

## Use case
- Faster profile scraping
- Better scraping output

PR #43, opened 2025-08-15
Feat 23 add fast linkedin option

qosha1:
Optional implementation of the fast linkedin scraper library with a configurable switch between selenium and playwright.

If interested @stickerdaniel  I can go back and clean it up to make it merge worthy but figured I'd share the raw one regardless.

RE: https://github.com/stickerdaniel/linkedin-mcp-server/issues/23

(btw - thanks for the MCP server, it took a bit to get going but it's pretty slick)

- Q

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ __pycache__/
 
 # C extensions
 *.so
-
+test-config.json
 # Distribution / packaging
 .Python
 build/
@@ -197,3 +197,4 @@ cython_debug/
 # claude code settings
 .claude
 CLAUDE.md
+compose/local/ngrok/ngrok.yml
--- a/compose/local/ngrok/Dockerfile
+++ b/compose/local/ngrok/Dockerfile
@@ -1,0 +1,12 @@
+FROM ngrok/ngrok:latest
+
+RUN ngrok --version
+
+# Add config script (if you want to set up multiple tunnels or use diff config)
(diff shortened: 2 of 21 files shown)


stickerdaniel:
Hey, really appreciate the effort you put into this. 
The factory pattern is solid work, but I'd like to propose a simpler direction
- I'll add native async support to fast-linkedin-scraper
- We'll migrate to exclusively using the new scraper and drop dual support (resolves #42)

This would avoid the threading complexity in the scraper adapter.

Thanks for pushing this forward!

qosha1:
Cool! Yea that would def make things easier lol I had to hard workaround the sync-async. 

Are you essentially deprecating the original linkedin-scraper? Eg. is the primary build path going forward the fast-linkedin-scraper? Cause if so I can start focusing on that too and likely help out on a few things as I build my workflow out. Would be good to be working on the one that's more likely to be kept up. 

Happy to push those as well should anything be of interest. 

Q

stickerdaniel:
Yes exactly, the plan is to remove the
original selenium scraper entirely as a
dependency.
While a few fixes were merged in the
original linkedin-scraper, it's become quite
outdated and the maintainer isn't very
active anymore. Fast-linkedin-scraper should
cover all the functionality from the
original library in the long term.
Feel free to jump in on the new repo.

I'll finish the async conversion today.

PR #310, opened 2026-04-02
fix(connect): sticky navbar viewport issue and multi-thread conversation bug

### LINKED ISSUE #304: [BUG] connect_with_person fails when Connect button is behind More menu (sticky navbar viewport issue)
## Problem

`connect_with_person` fails with a timeout error when the LinkedIn profile has the **Connect button hidden behind the More (•••) menu** — which happens for 3rd-degree connections.

The root cause is Playwright's click action failing because the element is reported as `outside of the viewport` even after `scroll_into_view_if_needed()`. This is caused by LinkedIn's **sticky navbar covering the element** after scrolling.

**Error log:**

```
- scrolling into view if needed
- done scrolling
- element is outside of the viewport
- retrying click action
```

## Root Cause

Two places in `extractor.py`:

1. **`_open_more_menu`** — clicks `More` button without first scrolling it into view:
   ```python
   await more_btn.first.click(timeout=5000)
   ```

2. **`click_button_by_text`** — no fallback when sticky navbar blocks the element from being considered "in viewport":
   ```python
   await target.click(timeout=timeout)
   ```

## Fix

**`_open_more_menu`** — add `scroll_into_view_if_needed` before click:
```python
await more_btn.first.scroll_into_view_if_needed(timeout=5000)
await more_btn.first.click(timeout=5000)
```

**`click_button_by_text`** — add `force=True` fallback:
```python
try:
    await target.click(timeout=timeout)
    return True
except Exception:
    pass
try:
    await target.click(timeout=timeout, force=True)
    return True
except Exception:
    return False
```

`force=True` bypasses Playwright's viewport visibility checks, which is appropriate here since LinkedIn's sticky header is the only thing preventing the click — the element itself is interactable.

## Tested

- Version: 4.8.1
- Profile type: 3rd-degree connection (Connect button under More menu)
- After patch: `status: connected`, `message: Connection request sent.` ✅

Happy to submit a PR if you'd like!

### LINKED ISSUE #307: [BUG] get_conversation fails when user has multiple threads with the same person
## Setup

uvx on macOS (arm64), VS Code Copilot Chat MCP client

## What Happened

When a user has multiple separate conversation threads with the same LinkedIn contact, `get_conversation` with `linkedin_username` fails to find any conversation, even though `search_conversations` correctly lists both threads in the sidebar.

I have two threads with the same contact:
- Thread A (Mar 17)
- Thread B (Apr 2)

`search_conversations(keywords="<contact name>")` shows both threads in the sidebar results. However:
- `get_conversation(linkedin_username="<contact-username>")` → "Could not find a conversation"
- `get_conversation(thread_id="<thread_id from search URL>")` → loads only the newer thread (Thread B); no way to retrieve the older thread (Thread A)

Expected: `get_conversation` should find at least one conversation by username, and there should be a way to access older threads when multiple exist with the same person.

## Steps to Reproduce

1. Have two separate message threads with the same LinkedIn contact (e.g., the contact sends a new InMail creating a second thread at a later date)
2. Call `search_conversations(keywords="<contact name>")` — both threads appear in the sidebar list with distinct subjects and dates
3. Call `get_conversation(linkedin_username="<contact-username>")` — returns "Could not find a conversation"
4. Call `get_conversation(thread_id="<thread_id_from_search_results_URL>")` — loads only the most recent thread; no way to get the older thread ID or its content

## Logs

Server log, trace JSONL, and generated issue template:
https://gist.github.com/dpearson2699/5a82c12261ba342b56f4aafec7ef641f

## Additional context

- The contact's profile URL slug differs from their current display name (name change after marriage), which may contribute to the username lookup failure.
- The core issue is the inability to access older threads when multiple exist with the same person. `search_conversations` proves both threads exist, but neither `get_conversation` path can reach the older one.
- Possible improvements: expose per-thread IDs from `search_conversations` sidebar results, or add an index/offset parameter when multiple threads exist for the same contact.

Gabrcodes:
## Summary

Fixes two reported bugs in `extractor.py`:

**1. `connect_with_person` fails when Connect is behind More menu (Closes #304)**

The root cause is Playwright reporting the element as "outside of the viewport" due to LinkedIn's sticky navbar covering it after scrolling.

- Added `scroll_into_view_if_needed` before clicking the More button
- Added `force=True` fallback when the sticky navbar still blocks the click
- Multi-selector fallback for the More button (4 selectors instead of 1)
- Same `force=True` fallback pattern added to `click_button_by_text`

**2. `get_conversation` fails with multiple threads per person (Closes #307)**

When a user's display name differs from their URL slug (e.g., name change after marriage), the inbox search can't find the conversation.

- `_open_conversation_by_username` now tries both the display name and the URL slug (hyphens replaced with spaces)
- Clear error message suggesting `thread_id` from `search_conversations` when lookup fails

## Test plan

- [x] 357 passed, 5 skipped, 0 failures
- [x] `ruff check` and `ruff format` pass
- [ ] Live test: connect with a 3rd-degree profile (Connect behind More menu)
- [ ] Live test: get_conversation with a contact whose name changed


hugobiais:
@Gabrcodes @stickerdaniel what's the update on this guys? I'm interested this PR is supposed to close #304 but it seems like it has gotten really big (making it more difficult to merge)

PR #339, opened 2026-04-09
added tools for posts and comments via the official api

tronicum:
needs app and user tokens. you might need to have an small biz (freelance should work) and the api calls are rate limited (100-500 p.d.)


stickerdaniel:
Re your LinkedIn qs:

**Flags:** rename `--linkedin-auth` → `--api-login`. Keep `--login` for the browser scrape session. Symmetrical, makes the two auth worlds obvious.

**ugcPosts → /rest/posts:** fold into this PR, don't want to ship the deprecated endpoint. Changes:
- `POST /rest/posts` / `DELETE /rest/posts/{urn}`
- Header `LinkedIn-Version: 202604` (const, bumpable)
- Flat body: `commentary`, `visibility`, `distribution: {feedDistribution: "MAIN_FEED", targetEntities: [], thirdPartyDistributionChannels: []}`, `lifecycleState: "PUBLISHED"`, `isReshareDisabledByAuthor: false` — drop the `specificContent.com.linkedin.ugc.ShareContent` wrapper
- Comments: `/rest/socialactions/{postUrn}/comments` with the same version header
- Scope / author URN / returned `x-restli-id` unchanged → tool return shapes stay

**Drop link share (`url` param on create_post).** URLs inline in `commentary` already auto-render with a preview card, and the new API requires caller-supplied title/description/thumbnail-URN (no OG scrape). Not worth the extra machinery for an agent-facing tool. Text + comments is the MVP; articles/images/reshare can come later if asked for.

PR #343, opened 2026-04-10
Add agent add install

pea3nut:
## Summary

Restructures the Installation section under uvx Setup to lead with a one-command [agent-add](https://github.com/pea3nut/agent-get) install, followed by manual JSON configuration as a fallback. agent-add auto-detects the user's AI host and writes the config — no need to find config file paths manually.

## What changed

- **Reordered**: agent-add command is now the primary install method under "### Installation"
- **Kept**: Manual JSON configuration moved under a "Manual Configuration" subsection
- **Kept**: All other installation methods (MCP Bundle, Docker, Local Setup) unchanged


pea3nut:
Thanks for the review! Both issues have been fixed:

PR #319, opened 2026-04-03
feat: add French locale support for connection state detection

vrpctaywal:
## Summary

- `connect_with_person()` always returns `connect_unavailable` for users whose LinkedIn interface is in French
- Root cause: `detect_connection_state()` only matches English button labels (`Connect`, `Follow`, `Pending`) but French LinkedIn displays `Se connecter`, `Suivre`, `En attente`
- The `_ACTION_AREA_END` regex only matches English section headings (`About`, `Experience`) but French displays `Infos`, `Expérience`, `Activité`

## Changes

- **`connection.py`**: `detect_connection_state()` now returns `(state, is_french)` tuple, checking both English and French button labels. Added `STATE_BUTTON_MAP_FR` and French section headings to `_ACTION_AREA_END`.
- **`extractor.py`**: `connect_with_person()` selects the correct button map based on detected language. `_open_more_menu()` now matches both `More`/`Plus` aria-labels and `Connect`/`Se connecter` menu items.
- **Tests**: Updated 8 existing tests for tuple return value + added 5 new French locale tests (all 14 pass).

## Test plan

- [x] All 14 `TestDetectConnectionState` tests pass
- [ ] Manual test: `connect_with_person` on a French LinkedIn profile with `Se connecter` button
- [ ] Manual test: `connect_with_person` on a French LinkedIn profile with `Suivre` + More menu

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #352, opened 2026-04-13
fix(connect): locale-agnostic connection detection (fixes German / supersedes #319)

Dominien:
## Summary

\`connect_with_person\` currently returns \`connect_unavailable\` on every profile for users whose LinkedIn UI is not in English. On my German account every invite fails with \`LinkedIn did not expose a usable Connect action for this profile.\` even though the profile clearly renders a \`Vernetzen\` button.

## Root cause

\`core/browser.py\` launches Playwright with \`locale=\"en-US\"\`. That option only drives \`navigator.language\` + the \`Accept-Language\` header — **LinkedIn ignores both for logged-in users** and serves the UI in the language chosen under *Settings → Account preferences → Display language* on the account itself. So a US-locale browser can still render:

| Account language | Connect | Follow | Pending | More |
|---|---|---|---|---|
| English (en) | Connect | Follow | Pending | More |
| German (de) | Vernetzen | Folgen | Ausstehend | Mehr |
| French (fr) | Se connecter | Suivre | En attente | Plus |

\`scraping/connection.py\` only matched the English strings, so every non-English profile fell through to \`unavailable\`.

## Changes

Kept \`locale=\"en-US\"\` as a best-effort hint and added small per-locale tables in \`scraping/connection.py\` so detection works regardless of what LinkedIn actually renders:

- \`STATE_BUTTON_MAP_BY_LOCALE\` — click labels keyed by locale + state. Ships \`en\`, \`de\`, \`fr\`. New locales = one line.
- \`_DETECTION_LABELS_BY_LOCALE\` — labels used while parsing the action area.
- \`_SECTION_HEADINGS_BY_LOCALE\` — per-locale section headings that end the action area (\`About\` / \`Info\` / \`Infos\` / …).
- \`_FIRST_DEGREE_MARKERS\` — \`· 1st\` / \`· 1.\` / \`· 1er\`.
- \`_MORE_ARIA_LABELS\` — \`More\` / \`Mehr\` / \`Plus\`.
- \`all_button_texts(state)\` helper so locale-agnostic call sites (like the More menu) can build their locator text lists.

\`detect_connection_state\` now returns \`tuple[ConnectionState, str]\` where the second element is the detected locale, so the caller picks the right button without needing its own language guess. \`connect_with_person\` looks up \`STATE_BUTTON_MAP_BY_LOCALE[locale][state]\`; \`_open_more_menu\` builds its aria-label selector and menu-item regex dynamically from the locale table.

## Why one PR for EN + DE + FR

#319 already proposed French support using an \`is_french: bool\` flag. That pattern doesn't generalize — adding German would require an \`is_german: bool\`, then Italian, Spanish, Dutch, etc. This PR introduces a single locale-keyed table so every new language is a one-line change in one file, and covers French as well to avoid leaving two overlapping PRs open. Credit to @vrpctaywal for pinning down the French case and the \`_contains\` helper shape.

## Tests

- Updated the 9 existing \`TestDetectConnectionState\` cases for the new tuple return value.
- Added 6 German cases (connectable, follow_only, pending, incoming_request, already_connected, action-area cut-at-\`Info\`).
- Added 5 French cases (parity with #319).
- **All 383 tests pass** (\`uv run pytest\`).
- \`uv run ruff check\`, \`uv run ruff format --check\`, \`uv run pre-commit run\` — all clean.

## Test plan

- [x] \`TestDetectConnectionState\` — 20 unit tests cover EN + DE + FR
- [x] Full test suite — 383 passed
- [ ] Manual: \`connect_with_person\` on a German profile with a \`Vernetzen\` button (verified locally — previously returned \`connect_unavailable\`, now sends the invite)
- [ ] Manual: \`connect_with_person\` on a French profile with \`Se connecter\`
- [ ] Manual: \`connect_with_person\` on a profile with only \`Folgen\` → goes through More menu


Dominien:
Closing this — #396 and #409 shipped a locale-independent detection that covers the same bug. Different technique (URL anchors vs. the per-locale string tables here) but the outcome lands for non-English LinkedIn accounts, which was the goal. Glad it's in. 👍

PR #366, opened 2026-04-15
verify connect flow before returning success & use custom invite deeplink

### LINKED ISSUE #365: [BUG] connect_with_person can return false-positive success
### Setup
Windows 11, local editable checkout with `uv run`, Claude/Codex-style MCP usage, real LinkedIn browser session.

### What Happened
`connect_with_person` could return a success response after clicking the UI without verifying that LinkedIn actually changed the relationship state. In practice this caused false positives where the tool reported a sent or accepted connection even though the profile did not reliably transition to `Pending` or `1st`.

### Steps to Reproduce
1. Run the server from a local checkout and authenticate with LinkedIn.
2. Call `connect_with_person` on a connectable profile.
3. Observe that the previous implementation inferred the pre-click state from the profile text, clicked the inferred action, and returned success without requiring a verified post-action state.
4. Compare the returned success with the actual profile state when LinkedIn does not transition to `Pending` or `1st`.

### Logs
Observed live behavior before the fix: the flow depended on inferred button clicks and optimistic success. The fixed version now verifies `Pending` after invite send and `1st` after accepting an incoming request.

yashkhare0:
## Summary

This fixes `connect_with_person` false positives by making the tool verify the final LinkedIn relationship state before returning success.

## What changed

- routes connectable and follow-only profiles through LinkedIn's `preload/custom-invite` deeplink
- prefers the explicit `Send without note` action when no note is provided
- re-reads the profile and only returns `connected` when the page verifies `Pending`
- re-reads the profile after accepting an incoming request and only returns `accepted` when the page verifies a `1st` connection
- adds test coverage for verified success and verification failure paths
- ignores local `/.serena` workspace artifacts in git

## Why

The previous flow inferred the pre-click state from profile text, clicked the UI, and could return success without proving that LinkedIn had actually transitioned to the expected final state. That made the connect tool flaky and could produce false-positive success responses.

## Validation

- `uv run pytest tests/test_scraping.py -k TestConnectWithPerson`
- `uv run ruff check linkedin_mcp_server/scraping/extractor.py tests/test_scraping.py`
- Live check on `drowsy` returned `pending` without sending a new invite
- Live check on `stanimirovicuros` sent an invite and verified the profile transitioned to `Pending`

Closes #365.

## Synthetic prompt

> Replace the optimistic `connect_with_person` flow with a verified sequence: short-circuit on already-connected or pending states, send invites through LinkedIn's `preload/custom-invite` deeplink, and only return success after re-reading the profile and confirming `Pending` (or `1st` after accepting an incoming request). Update the tests to enforce verification and ignore the local `.serena` workspace artifacts.

Generated with Codex (GPT-5)


stickerdaniel:
Thanks for the PR. The verify-after-action layer is exactly what #365 needs and I'd like to land it.

I'm on board with the deeplink approach. Looking at how LinkedIn actually emits `/preload/custom-invite/` as the Connect anchor's href in its own HTML, navigating to it directly drops the whole button-click + modal-index dance that #304, #319, #348 and #368 all orbited, and makes the Connect flow locale-independent by default.

Before we merge:

1. Rebase on main, #368 rewrote the block you're editing.
2. Drop `_click_send_without_note_button`. Use `_click_dialog_primary_button()` (positional, from #368). The regex on "Send without a note" is English-only and LinkedIn localizes the label, which defeats the locale advantage the deeplink gives us.
3. Drop the on-profile click fallback. If the deeplink doesn't render a dialog, return `connect_unavailable`. One reliable path is easier to reason about than two.
4. Keep Accept as on-profile click for `incoming_request`. It's not the fragile path, no modal involved.
5. Move `.serena` to your global gitignore, not the project one.

Happy to re-review after the rebase.

hugobiais:
Hi @yashkhare0 — since this PR has been quiet for ~9 days I went ahead and rebased it onto current `main` per @stickerdaniel's review, just to keep things moving. Branch: https://github.com/hugobiais/linkedin-mcp-server/tree/feature/366-verified-connect

Applied changes from the review:

1. Rebased on current `main` (which now includes #368).
2. Dropped `_click_send_without_note_button` — both the note and noteless paths in `_send_connection_invite_via_deeplink` now go through `_click_dialog_primary_button()` (positional, locale-independent).
3. Dropped the on-profile click fallback for `connectable` / `follow_only`. If the deeplink dialog doesn't open, returns `connect_unavailable`.
4. Kept `incoming_request` as on-profile click (no modal).
5. Did not include the `.serena` `.gitignore` change.

Verification:

- `uv run ruff check .` — clean
- `uv run ty check linkedin_mcp_server/scraping/extractor.py` — clean
- `uv run pytest tests/test_scraping.py -k TestConnectWithPerson` — 12 passed (incl. `test_returns_send_failed_when_pending_verification_fails`)

Feel free to `git fetch` from my fork and pull this into your branch, or just cherry-pick the single commit (`46d914d`). Happy to open a separate PR instead if that's easier — let me know.

PR #388, opened 2026-04-23
feat: detect #OpenToWork badge from profile scrapes

easternbloc:
## Summary

- Intercepts LinkedIn's `identityDashOpenToCardsByTopCard` Voyager API response (fired automatically when loading a profile page) to detect whether a profile has the #OpenToWork badge enabled
- Adds a passive `page.on("response")` listener in `scrape_person()` with proper cleanup via `try/finally`
- Surfaces the result as an `open_to_work` boolean field in the `scrape_person()` return dict

## Implementation

The listener checks for `OpenToCards` in the response URL and parses the JSON payload to extract card elements. If elements are present, the profile has OpenToWork enabled. The listener is registered before the section scraping loop and removed in a `finally` block to avoid leaks on error paths.

No new dependencies — uses the existing `page.on("response")` pattern already established in the codebase.

## Test plan

- [x] Module imports cleanly (`from linkedin_mcp_server.scraping.extractor import LinkedInExtractor`)
- [x] All 384 existing tests pass (1 pre-existing failure in `test_browser_security.py` unrelated to this change)
- [x] Ruff lint passes with no warnings
- [ ] Manual verification: scrape a profile with #OpenToWork enabled and confirm `open_to_work: true` in the result
- [ ] Manual verification: scrape a profile without #OpenToWork and confirm `open_to_work: false` or field absent

### DIFF
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -237,7 +237,7 @@ class LinkedInExtractor:
             else:
                 self._open_to_work_data = {"open_to_work": False}
         except Exception:
-            pass
+            logger.debug("OpenToWork API capture failed for %s", url)
 
     @staticmethod
     def _normalize_body_marker(value: Any) -> str:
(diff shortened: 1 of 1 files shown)


easternbloc:
Closing — the implementation needs rethinking to cover OTW detection in search results, not just individual profile pages. Will revisit with a more complete approach.

PR #413, opened 2026-05-02
fix(messaging): expose per-thread refs in search and add index for multi-thread participants

### LINKED ISSUE #307: [BUG] get_conversation fails when user has multiple threads with the same person

stickerdaniel:
## Summary

LinkedIn renders the messaging sidebar with no anchor `href`s, no `data-thread-*` attributes, and no embedded URNs (verified via DOM probe), so callers had no way to discover thread IDs from search results and could not reach an older thread when one participant has multiple conversations. Issue #307 reports this as: search shows both threads, but `get_conversation(linkedin_username=...)` returns "Could not find a conversation" and the thread ID copied out of the search URL points to the wrong (auto-loaded most-recent) thread.

- `search_conversations` now drives search via the `?searchTerm=` URL parameter (no more flaky searchbox typing) and click-to-captures per-result thread URLs as `kind=conversation` references — same shape `get_inbox` already uses.
- `_resolve_conversation_thread_urls` returns every thread whose participant `aria-label` matches the profile display name *exactly*, replacing the substring matcher that broke on cached old names after a profile rename.
- `get_conversation` accepts a 0-based `index` parameter to disambiguate when one participant has multiple threads; out-of-range raises with the actual count.
- The shared click-to-capture helper waits on label `state="attached"` instead of the default `visible` (Ember-managed labels never satisfy Playwright's visibility heuristic) and polls for the SPA URL change instead of a fixed 300 ms sleep that raced LinkedIn's click-handler binding — both fixed long-standing flakes seen with `get_inbox`.

Resolves: #307

## Test plan

- [x] `uv run pytest` — 436 passed
- [x] `uv run ruff check .` and `uv run ty check` — clean
- [x] Live verification against LinkedIn:
  - `search_conversations(keywords="Daniel")` → 11 conversation refs, each with a unique thread URL (was 0 before, plus a "search input not found" warning)
  - `get_conversation(linkedin_username="fadi-eliwi")` → loads correct thread with default `index=0`
  - `get_conversation(linkedin_username="fadi-eliwi", index=99)` → raises `index 99 out of range: only 1 thread(s) exist for fadi-eliwi.`

### DIFF
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -2541,6 +2541,13 @@ class LinkedInExtractor:
         InMail. Ignored when ``thread_id`` is provided. Use
         ``search_conversations`` to enumerate thread IDs first if disambiguation
         by index is impractical.
+
+        Side effect when looked up by username: resolution searches LinkedIn's
+        messaging inbox for the participant's display name and click-visits
+        every matching row to capture its thread ID (no anchor hrefs or
+        thread-id attributes exist in the sidebar). Each visit selects the row
+        in the LinkedIn UI and may mark it as read. Pass ``thread_id`` directly
+        to skip this enumeration.
         """
         if not linkedin_username and not thread_id:
             raise LinkedInScraperException(
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -1831,102 +1831,51 @@ class LinkedInExtractor:
         match = re.search(r"/messaging/thread/([^/?#]+)/", url)
         return match.group(1) if match else None
 
-    async def _resolve_conversation_thread_url(self, search_query: str) -> str | None:
-        """Search the messaging inbox and return the matching thread URL."""
-        await self._navigate_to_page("https://www.linkedin.com/messaging/")
+    async def _resolve_conversation_thread_urls(self, display_name: str) -> list[str]:
+        """Return all thread URLs whose participant aria-label matches display_name.
+
+        Uses URL-driven search (`/messaging/?searchTerm=…`) plus click-to-capture
+        because LinkedIn renders the messaging sidebar with no anchor hrefs, no
+        data-thread attributes, and no embedded URNs — clicking each row and
+        reading the resulting SPA URL is the only available extraction path.
+
+        Matching is by exact ``aria-label == "Select conversation with <name>"``,
+        which is locale-independent and tolerates duplicate threads with the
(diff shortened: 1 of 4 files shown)


stickerdaniel:
* **#413** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/413?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/413?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

PR #447, opened 2026-05-15
feat(network): add get_pending_invitations tool

### LINKED ISSUE #446: [FEATURE] Add get_pending_invitations tool for network invitations
### Feature Description

Add a new MCP tool `get_pending_invitations` that lists pending LinkedIn network invitations (received or sent), mirroring the page at `https://www.linkedin.com/mynetwork/invitation-manager/`.

Initial scope is read-only. Accept / ignore / withdraw actions are intentionally out of scope for the first iteration to keep the PR focused; they can land in a follow-up if this one is accepted.

### Use Case

The current toolset covers messaging (`get_inbox`, `get_conversation`) but exposes no surface for pending connection invitations, which is a primary inbound channel on LinkedIn:

- Recruiters frequently use "Invite + note" instead of InMail. Those messages live in `/mynetwork/invitation-manager/` and never appear in `get_inbox` until the invitation is accepted, so an AI agent assisting with inbound triage cannot see them today.
- Users need a non-destructive way to triage incoming invitations before deciding whether to accept (which would then make the conversation visible to other tools in this server).
- Symmetric `sent` listing is useful for reviewing outstanding outbound invitations a user has issued.

### Suggested Approach

Follow the existing `get_inbox` pattern very closely:

- New file `linkedin_mcp_server/tools/network.py` exposing `register_network_tools(mcp, *, tool_timeout)` and the tool function `get_pending_invitations(ctx, limit, kind, extractor)`.
- New extractor method `LinkedInExtractor.get_pending_invitations(limit, kind)` in `scraping/extractor.py`. Navigation -> `detect_rate_limit` -> `_wait_for_main_text` -> `handle_modal_close` -> `_scroll_main_scrollable_region` -> `_extract_root_content(["main"])` -> `strip_linkedin_noise` -> `build_references` -> `_single_section_result`, identical to `get_inbox`.
- Register the new module from `server.py::create_mcp_server()`.
- Parameter shape: `limit: int` (1-100, default 20), `kind: Literal["received", "sent"] = "received"`. URL: `/mynetwork/invitation-manager/{kind}/`.
- Return format: standard `{url, sections: {invitations: raw_text}, references?: {invitations: [...]}}`. No new section dict in `fields.py` needed — this is a one-page tool, not a sectioned scraper.
- Tests: `TestNetworkTools` class in `tests/test_tools.py` mirroring `TestMessagingTools::test_get_inbox_success`; add `get_pending_invitations` to `_make_mock_extractor`.
- Docs: README tool table, `docs/docker-hub.md` features list, `manifest.json` tools array.

I'm happy to open the PR myself; filing this issue first per `CONTRIBUTING.md`.

ghul0:
## Summary

Adds a new read-only MCP tool `get_pending_invitations` for listing pending LinkedIn network invitations (received or sent). This fills the gap noted in #446 — recruiter "Invite + note" messages live in `/mynetwork/invitation-manager/` and never reach `get_inbox` until accepted, so an agent assisting with inbound triage cannot see them today.

**What's in scope**

- New tool exposed via `tools/network.py::register_network_tools` with signature `get_pending_invitations(ctx, limit=20, kind="received"|"sent")`.
- New extractor method `LinkedInExtractor.get_pending_invitations(limit, kind)` that follows the `get_inbox` template line-for-line: `_navigate_to_page` -> `detect_rate_limit` -> `_wait_for_main_text` -> `handle_modal_close` -> `_scroll_main_scrollable_region` -> `_expand_invitation_note_toggles` -> `_extract_root_content(["main"])` -> `strip_linkedin_noise` -> `build_references` -> `_single_section_result`.
- Auto-expansion of truncated invitation notes via the new `_expand_invitation_note_toggles` helper. Uses the locale-independent `data-testid="expandable-text-button"` selector, injects a scoped one-shot stylesheet to re-enable `pointer-events` (LinkedIn ships these buttons with inline `pointer-events: none`), and dispatches a synthetic bubbling MouseEvent so the React handler fires. A button is tagged with `data-mcp-clicked` only **after** a successful dispatch, so a click that throws is left unmarked and retried on the second pass; the selector also skips `aria-expanded="true"` so an already-expanded note is never toggled closed.
- Empty-state handling for `kind="received"`: when the selected ALL-count tab reads zero, the method returns an empty single-section result instead of scraping the unrelated "people you may know" recommendation cards LinkedIn renders below the empty invitations list. The zero check is locale-independent — it locates the selected tab via `aria-current="true"` + the stable `/invitation-manager/received/ALL` URL, and only the digit count is read as text (guarded behind that structural selector, documented in the helper docstring).
- Per-call `limit` bounds the returned references (`build_references(...)[:limit]`), and `_REFERENCE_CAPS["invitations"]` is set to 100 to match the tool's `Field(le=100)` ceiling so larger invitation lists are not silently truncated at the default cap of 12.
- Registration in `server.py::create_mcp_server()`.
- Tests: `TestNetworkTools` in `tests/test_tools.py` (happy path, `sent` kind, invalid kind rejection, limit-bound rejection; `_make_mock_extractor` extended; tool added to both timeout-coverage tuples) plus `TestGetPendingInvitations` in `tests/test_scraping.py` (extractor-level regression coverage: `[:limit]` reference trimming, the `sent` navigation target, the `aria-expanded="true"` / `data-mcp-clicked` skip guards, and the zero-count empty-state path).
- README tool table, `docs/docker-hub.md` features list, and `manifest.json` tools array all updated.

**Intentionally out of scope**

- Accept / ignore / withdraw actions are deferred to a follow-up PR if this lands. Keeping the surface area read-only here both shrinks the diff and avoids exposing destructive operations on a surface that hasn't been exercised yet.
- No new entry in `fields.py` — this is a one-page tool, not a sectioned scraper, matching `get_feed` / `get_inbox`.

## Testing

- `uv run pytest`: 551 passed locally. One pre-existing failure in `test_browser_security.py::test_harden_linkedin_tree_noop_outside_linkedin` reproduces on `main` before this branch with `umask=002`; it's environmental and unrelated.
- `uv run ruff check .`: clean. `uv run ty check`: clean. `uv run pre-commit run --all-files`: all hooks pass.
- Live verified end-to-end via `get_pending_invitations` (read-only) on a real account in two states:
  - non-empty `kind="received"`: result shape conforms to `{url, sections, references}`; the section text contained inviter blocks with every truncated invitation note auto-expanded (`text` matched the collapse verb N times, the expand verb 0 times); references carried `kind: "person"` entries with `/in/USERNAME/` URLs, sliced to `limit`.
  - zero `kind="received"`: returns `sections: {}` with no references, confirming the "people you may know" recommendation cards are not surfaced as invitations.

## Synthetic prompt

> Add a read-only `get_pending_invitations` MCP tool that lists pending LinkedIn network invitations (received or sent) by navigating to `/mynetwork/invitation-manager/{kind}/`, mirroring the existing `get_inbox` extractor pattern. Before extracting, auto-expand truncated invitation notes by dispatching a synthetic click on each `data-testid="expandable-text-button"` (inject a scoped `pointer-events` stylesheet override, mark each button `data-mcp-clicked` only after a successful dispatch so failed clicks retry, and skip `aria-expanded="true"` so expanded notes are never re-collapsed). Bound returned references to the per-call `limit` and add an `invitations` entry to `_REFERENCE_CAPS`. For `kind="received"`, when the selected ALL-count tab reads zero, return an empty result instead of scraping the "people you may know" recommendations, locating the tab via the locale-independent `aria-current="true"` + `/received/ALL` URL signal. Wire it through `tools/network.py` and `server.py::create_mcp_server()`, add `TestNetworkTools` in `tests/test_tools.py` and `TestGetPendingInvitations` in `tests/test_scraping.py`, and update README, `docs/docker-hub.md`, and `manifest.json`. Accept/ignore/withdraw actions are intentionally out of scope.

Generated with Claude Opus 4.7 and GPT-5.5

Closes #446

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ Through this LinkedIn MCP server, AI assistants like Claude can connect to your 
 | `get_conversation` | Read a specific messaging conversation by username or thread ID | [#434](https://github.com/stickerdaniel/linkedin-mcp-server/issues/434) |
 | `search_conversations` | Search messages by keyword | working |
 | `send_message` | Send a message to a LinkedIn user (requires confirmation) | [#433](https://github.com/stickerdaniel/linkedin-mcp-server/issues/433) [#441](https://github.com/stickerdaniel/linkedin-mcp-server/issues/441) |
+| `get_pending_invitations` | List pending network invitations (received or sent) from `/mynetwork/invitation-manager/` | working |
 | `get_company_profile` | Extract company information with explicit section selection (posts, jobs); about-section references may include a `company_urn` entry carrying the numeric id used by LinkedIn's people-search `currentCompany` URL facet | working |
 | `get_company_posts` | Get recent posts from a company's LinkedIn feed | working |
 | `search_companies` | Search for companies on LinkedIn by keywords | working |
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ This MCP server is **free** and **open source**, supported by [**Unipile**](http
 | `get_conversation` | Read a specific messaging conversation by username or thread ID | working |
 | `search_conversations` | Search messages by keyword | working |
 | `send_message` | Send a message to a LinkedIn user (requires confirmation) | [#433](https://github.com/stickerdaniel/linkedin-mcp-server/issues/433) [#441](https://github.com/stickerdaniel/linkedin-mcp-server/issues/441) [#483](https://github.com/stickerdaniel/linkedin-mcp-server/issues/483) [#560](https://github.com/stickerdaniel/linkedin-mcp-server/issues/560) |
+| `get_pending_invitations` | List pending network invitations (received or sent) from `/mynetwork/invitation-manager/` | working |
 | `get_company_profile` | Extract company information with explicit section selection (posts, jobs); about-section references may include a `company_urn` entry carrying the numeric id used by LinkedIn's people-search `currentCompany` URL facet | working |
 | `get_company_posts` | Get recent posts from a company's LinkedIn feed | working |
 | `search_companies` | Search for companies on LinkedIn by keywords | working |
--- a/docs/docker-hub.md
+++ b/docs/docker-hub.md
@@ -9,6 +9,7 @@ A Model Context Protocol (MCP) server that connects AI assistants to LinkedIn. A
 - **Profile Access**: Get detailed LinkedIn profile information including experience, education, skills, projects, certifications, and more
 - **Own Profile**: Fetch the authenticated user's own profile to give agents self-context
 - **Profile Connections**: Send connection requests or accept incoming ones, with optional notes
+- **Network Invitations**: List pending network invitations (received or sent) from `/mynetwork/invitation-manager/`
 - **Company Profiles**: Extract comprehensive company data, including the LinkedIn company URN id (used by LinkedIn's people-search `currentCompany` URL facet)
 - **Company Employees**: List employees at a company with optional keyword filtering
 - **Company Search**: Search for companies by keyword
--- a/docs/docker-hub.md
+++ b/docs/docker-hub.md
@@ -7,6 +7,7 @@ A Model Context Protocol (MCP) server that connects AI assistants to LinkedIn. A
 - **Profile Access**: Get detailed LinkedIn profile information including experience, education, skills, projects, certifications, and more
 - **Own Profile**: Fetch the authenticated user's own profile to give agents self-context
 - **Profile Connections**: Send connection requests or accept incoming ones, with optional notes
+- **Network Invitations**: List pending network invitations (received or sent) from `/mynetwork/invitation-manager/`
 - **Company Profiles**: Extract comprehensive company data, including the LinkedIn company URN id (used by LinkedIn's people-search `currentCompany` URL facet)
 - **Company Employees**: List employees at a company with optional keyword filtering
 - **Company Search**: Search for companies by keyword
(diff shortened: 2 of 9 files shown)


ghul0:
Thanks for the careful review — all three P1 items addressed in 3920cb1.

**1. Limit not enforced.** References are now sliced to `[:limit]` after `build_references`, so a `limit=5` call cannot return more than five inviter references. The `sections` text can still include slightly more cards because LinkedIn lazy-loads invitations a screenful at a time; the docstring now states this explicitly so callers don't get a surprise.

**2. References truncated at the default cap of 12.** Added an `invitations` entry to `_REFERENCE_CAPS` with a value of 100, matching the `Field(le=100)` ceiling on the tool. The per-call slice from item 1 is the operative ceiling in practice; the cap entry just removes the silent 12-link truncation at the `build_references` layer.

**3. Expanded notes could collapse.** Strengthened the expand-toggle selector to `[data-testid="expandable-text-button"]:not([aria-expanded="true"]):not([data-mcp-clicked])`. One nuance versus the inline suggestion: on this surface LinkedIn ships collapsed toggles **without** an `aria-expanded` attribute at all (verified during the original DOM debug pass — the inline suggestion used `[aria-expanded="false"]` which matches zero buttons in that state). The `:not([aria-expanded="true"])` form is the conservative shape that matches both the unset and the explicit `"false"` states while skipping the expanded ones.

CI green on the new commit; full suite passes locally aside from one pre-existing umask-sensitive case in `test_browser_security.py` that reproduces on `main` unchanged.

PR #449, opened 2026-05-16
fix(person): detect custom note quota block in connect_with_person

### LINKED ISSUE #448: [BUG] connect_with_person reports note_sent=true and send_failed when free invite-note limit is reached
### Setup

uv-managed local checkout (Linux, Python 3.13), driven from Claude Code via the stdio MCP transport. Behaviour reproduces on `main` at `e6430e2` — no local patches in the affected code path.

### What Happened

`connect_with_person(username=..., note=...)` returned

```json
{
  "status": "send_failed",
  "message": "Submitted the invite dialog but the profile still exposes Connect.",
  "note_sent": true
}
```

The actual LinkedIn-side state, verified by replaying the same flow with Playwright in `--no-headless` mode, was that LinkedIn refused the invite and rendered a Premium upgrade modal:

> **Wysyłaj nieograniczoną liczbę spersonalizowanych zaproszeń dzięki Premium.**
> Wyczerpałeś(-aś) limit bezpłatnych spersonalizowanych notatek...

i.e. the account had hit the free quota for **personalized** invitation notes. A note-less invite would have gone through; an invite with a note could not.

Two problems with the current response shape:

1. **`note_sent: true` is misleading.** `_submit_invite_dialog` sets it as soon as the textarea is filled (`extractor.py:1581`), regardless of whether the subsequent primary-button submit actually delivered the invite. When LinkedIn intercepts the submit with the Premium upsell, the textarea was filled but the note was *not* sent — yet the field still reports `true`.
2. **`status: send_failed` + the generic message don't tell the caller why.** A quota-blocked invite, a CAPTCHA, and a hung dialog all surface the same code. An LLM consuming the result has no way to recover (e.g. retry without a note) without scraping the modal text itself.

Suggested separate signal:

```json
{
  "status": "custom_note_limit_reached",
  "message": "LinkedIn blocked the invite — the free personalized invitation note limit has been reached.",
  "note_sent": false,
  "can_send_without_note": true
}
```

### Steps to Reproduce

1. Use an account whose free personalized-note quota for the current rolling window is already exhausted.
2. Call `connect_with_person(linkedin_username="<a stranger you can still invite>", note="any text")`.
3. Observe the response shape: `send_failed` + `note_sent: true`.
4. Replay the same flow against the same profile with the browser visible (`uv run -m linkedin_mcp_server --no-headless`) — LinkedIn renders the Premium upsell modal on submit instead of sending.

### Logs

```
INFO  Connection signals for <user>: state=connectable signals=ActionSignals(has_invite_anchor=True, ...)
INFO  Dialog opened via deeplink
INFO  Note textarea filled
INFO  Primary button clicked
INFO  Dialog closed (Premium modal opened instead — not detected by current code path)
INFO  Verify: has_invite_anchor=True → send_failed
```

### Root-Cause Pointer (for whoever picks this up)

- `extractor.py:1547-1613` `_submit_invite_dialog` — `note_sent` is set inside the textarea-fill branch and never re-evaluated after the primary-button click. It should track *delivered* state, not *filled* state.
- `extractor.py:1757-1764` — the `verified_signals.has_invite_anchor` branch is the only failure surface; a quota-block is currently funnelled through it without discrimination.

A locale-independent detector for the Premium upsell modal is the right wedge: the upsell renders a fresh `<dialog>` whose primary action carries an `href` pointing to `/premium/...` — that URL fragment is stable and language-independent, mirroring the existing `/preload/custom-invite/`-based gates in this codebase. When that signal fires, the result should set `note_sent: false`, surface a distinct status (e.g. `custom_note_limit_reached`), and optionally hint that a note-less retry is possible.

ghul0:
## Summary

Fixes #448. When LinkedIn rejects an invite-with-note because the account's free personalized-note quota is exhausted, the old code path could return misleading/generic states such as `send_failed` or `connect_unavailable`.

This PR keeps the response simple:

```json
{
  "status": "custom_note_limit_reached",
  "message": "<raw LinkedIn Premium dialog text>",
  "note_sent": false
}
```

No retry hints, no synthesized explanation. The `message` is read directly from the LinkedIn Premium upsell dialog.

**Behaviour changes**

- New status `custom_note_limit_reached` on `connect_with_person` with `note_sent: false`.
- Removed `can_send_without_note`; callers only get the status plus LinkedIn's own message text.
- `note_sent` now tracks delivery of the note, not textarea fill. Any failure path leaves it `false`.
- Existing statuses (`connected`, `accepted`, `send_failed`, `connect_unavailable`, etc.) are unchanged.

**Detection signal**

Locale-independent: the upsell renders a dialog whose body links to `/premium/...`. The detector waits for `dialog[open] a[href*="/premium/"]` or `[role="dialog"] a[href*="/premium/"]`, then reads the surrounding dialog text directly from LinkedIn.

**Implementation**

- `_get_premium_upsell_message` waits for the Premium link and returns the raw dialog `innerText` / `textContent`.
- `_submit_invite_dialog` returns `(submitted, note_sent, note_limit_message)`.
- Detection covers:
  1. `Add note` opening the Premium upsell before a textarea appears,
  2. filling the note failing because the upsell replaced the editor,
  3. submit being intercepted by the upsell.
- `_probe_invite_note_limit` handles no-anchor profiles by opening the deeplink only as a non-submitting probe and returning the raw dialog text if LinkedIn shows the upsell. It never clicks Send.
- Public `tools/person.py` docstring updated.

## Testing

Latest head: `d11bda3`.

- GitHub CI pending/running after latest push.
- Local: `uv run --python 3.13.9 ruff check .` clean.
- Local: `uv run --python 3.13.9 ruff format --check .` clean.
- Local: `uv run --python 3.13.9 ty check` clean.
- Targeted local tests: `34 passed` for `tests/test_scraping.py::TestConnectWithPerson` and `tests/test_tools.py::TestPersonTool`.
- Local integration branch with PR #447 merged: `72 passed` for connect/person, network tool, and link metadata coverage.
- Full local suite still has one pre-existing environment-sensitive failure in `tests/test_browser_security.py::test_harden_linkedin_tree_noop_outside_linkedin` on this machine (`umask=002` parent-dir mode); unaffected suite passes when that single known test is deselected.

## Synthetic prompt

> When `connect_with_person` hits LinkedIn's Premium upsell because personalized invite-note quota is exhausted, return `status: custom_note_limit_reached`, `note_sent: false`, and set `message` to the raw dialog text from LinkedIn. Do not add retry hints such as `can_send_without_note`. Detect the upsell via a locale-independent `/premium/` link in the dialog, read the surrounding dialog text directly, cover Add-note/fill/submit interception cases, and keep the no-anchor deeplink path probe-only so it never submits an invite.

Generated with Claude Opus 4.7

Closes #448


stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

PR #456, opened 2026-05-20
fix(connect): match alertdialog/aria-modal and wait for late-mounting invite modal

### LINKED ISSUE #455: connect_with_person fails when `note` is provided: new two-step 'Add a note?' gating dialog not handled (v4.13.0)
## Summary
On v4.13.0, `connect_with_person` with a `note` parameter consistently returns `connect_unavailable` ("LinkedIn did not open a usable invite dialog"). The root cause is a new LinkedIn UI variant: a two-step "gating" dialog that the current note path doesn't traverse.

The no-note path still works — see Workaround below.

## Environment
- linkedin-scraper-mcp **v4.13.0**, FastMCP 3.3.1
- Headless Patchright (default config)
- LinkedIn account in the UK; observed across mixed targets (2nd- and 3rd-degree, Engineering Manager through CTO). Repro is account-wide, not profile-specific.
- **2026-05-19:** 3/3 successful with-note connects on the same account.
- **2026-05-20:** 8/8 with-note attempts fail with `connect_unavailable`; 4/4 no-note attempts succeed on the same browser session. Suggests LinkedIn rolled the gating-dialog variant to this account overnight.

## What LinkedIn now serves
After `_navigate_to_page` to `/preload/custom-invite/?vanityName=<user>`, the *first* rendered dialog is a chooser — not the note textarea:

> **Add a note to your invitation?**
> Personalize your invitation to *X* by adding a note. LinkedIn members are more likely to accept invitations that include a note.
> `[ Add a note ]`   `[ Send without a note ]`

The textarea is only mounted after clicking **"Add a note"**. The current note path waits for the textarea directly, times out at 5s, and bails.

Screenshot from `~/.linkedin-mcp/trace-runs/<run-id>/screens/005-extractor-after-goto.png` — happy to attach if helpful.

## Reproduction
1. v4.13.0, logged-in profile.
2. `connect_with_person(linkedin_username="<2nd-degree target>", note="hi")`.
3. Returns `{"status": "connect_unavailable", "note_sent": false, ...}`.

## Diagnostic evidence (`--log-level DEBUG`, abridged)
```
scraping.extractor - INFO  - Connection signals for X: state=connectable signals=ActionSignals(has_invite_anchor=True, has_compose_anchor_in_action_root=True, has_edit_intro_anchor=False, has_labeled_action_button=True, has_labeled_action_anchor=True)
scraping.extractor - DEBUG - _navigate_to_page: target=https://www.linkedin.com/preload/custom-invite/?vanityName=X
scraping.extractor - DEBUG - Note textarea did not appear
```

Trace `body_marker` proves the gating dialog is open and rendered:
`"Dialog content start. Add a note to your invitation? Personalize your invitation to X by adding a note. LinkedIn members are more likely to accept invitations that ..."`

## Workaround
Call `connect_with_person` **without** the `note` parameter — the no-note path correctly clicks "Send without a note". Verified 4/4 successful no-note sends (2nd-deg IC, 2nd-deg CTO, 3rd-deg CTO) on the same session that just produced 8/8 with-note failures.

## Suggested fix
On the invite preload page, branch on what's actually in the DOM:
1. If the textarea is present immediately → legacy single-step flow → proceed as today.
2. Else, look for the chooser dialog (text contains "Add a note to your invitation") and click the **"Add a note"** button, then wait for the textarea to mount.

## Note on related issues
Distinct from existing reports:
- #454 — same `connect_unavailable` symptom but on the **no-note** path (reporter explicitly ruled out the note flow). May indicate LinkedIn is A/B-testing the gating dialog inconsistently across accounts/regions.
- #432 — selector collision on `.nth(22)` for the Emoji button; different code path.
- #407 — restricted-invite profiles where Send is disabled.

ivangotti:
Closes #455.

## TL;DR

The first cut of this PR (commit `a48fccc`) lowered the secondary-button guard from `btn_count >= 3` to `btn_count >= 2` to handle the new "Add a note to your invitation?" gating dialog. That change was correct *for the case where the dialog actually opens*, but on real accounts post-deploy `connect_with_person` was still returning `connect_unavailable` for every with-note attempt — because `_dialog_is_open` was failing one layer upstream and the secondary-button branch never ran.

This commit fixes the upstream failure. The 2-button-gating handling is preserved.

## Real root cause

Two distinct issues in `_dialog_is_open` / `_DIALOG_SELECTOR`, both visible in trace `run-8fic76qw` from a failing 8/8 production run on 2026-05-20 (post-`a48fccc` deploy):

1. **Selector was too narrow.** The gating dialog is an `artdeco-modal` whose root carries `role="alertdialog"` and/or `aria-modal="true"`, neither of which matched `dialog[open], [role="dialog"]`. Trace screenshots show the dialog visibly mounted at the exact DOM moment `_dialog_is_open` was returning False — the helper just couldn't see it.

2. **Race against modal mount.** `_dialog_is_open` short-circuited on `count() == 0` before spending its timeout budget. After `goto(/preload/custom-invite/?vanityName=)` resolves on `domcontentloaded`, the modal mounts a few hundred ms later, so the count check fired against an empty DOM and returned False immediately. `wait_for(state="visible")` is the right primitive — it polls until the element exists *and* is visible.

The two effects compound: even on accounts where LinkedIn picks `role="dialog"` and the selector would have matched, the eager `count()` made the helper racy; even on a slow account where waiting longer would have helped, the wrong selector would still miss the alertdialog mount.

## Evidence

From the trace (`run-8fic76qw/trace.jsonl`, screens `005`, `013`, `017`, `021`):

* Step 5: `goto preload/custom-invite/?vanityName=meenal-ganesh-b1710321`. Screenshot shows the gating dialog **fully mounted and visible** (X close + "Add a note" + "Send without a note"). Body marker: `"Dialog content start. Add a note to your invitation? Personalize your invitation to Meenal Ganesh by adding a note."`
* The dialog is structurally identical across Brian W., Igor Dimoski, Sandro Valentini, Greg Sandzimier, etc.
* For Greg (run-3mu94g3t, May 19), the trace shows a *verification scrape immediately after* the deeplink — i.e. `submitted=True`, dialog detected. Greg's profile after: `"More Pending Message"` — invite landed.
* For Brian / Igor / Sandro (run-8fic76qw, May 20), there is **no verification scrape after the deeplink** — the trace jumps straight to the next person's profile. `submitted=False` → `connect_unavailable`. Same dialog, different `_dialog_is_open` result.

The only thing that differed between the working and failing flows was a slight DOM/timing difference in the modal's mount path, which a single broadened selector + a real `wait_for` reliably absorbs.

## The fix

```diff
 _DIALOG_SELECTOR = (
+    'dialog[open], '
+    '[role="dialog"], '
+    '[role="alertdialog"], '
+    '[aria-modal="true"]'
 )

 _DIALOG_TEXTAREA_SELECTOR = (
+    '[role="dialog"] textarea, '
+    '[role="alertdialog"] textarea, '
+    '[aria-modal="true"] textarea, '
+    'dialog textarea'
 )

 async def _dialog_is_open(self, *, timeout: int = 1000) -> bool:
-    locator = self._page.locator(_DIALOG_SELECTOR)
+    locator = self._page.locator(_DIALOG_SELECTOR).first
     try:
-        if await locator.count() == 0:
-            return False
-        await locator.first.wait_for(state="visible", timeout=timeout)
+        await locator.wait_for(state="visible", timeout=timeout)
         return True
     except Exception:
         return False
```

Per-call sites: every existing reference to `_DIALOG_SELECTOR` and `_DIALOG_TEXTAREA_SELECTOR` (button collection, textarea fill, dismiss-on-hidden) gets the broadened scope automatically.

The 2-button-gating handling from `a48fccc` is preserved verbatim — with the dialog now correctly detected, the existing `btn_count >= 2` / `nth(btn_count - 2)` branch fires and reveals the textarea on the new gating layout, while the no-note path still hits "Send without a note" via `_click_dialog_primary_button`.

## Tests

* New `test_dialog_selector_matches_alertdialog_and_aria_modal` — guards every required ARIA pattern in `_DIALOG_SELECTOR` and `_DIALOG_TEXTAREA_SELECTOR`. Future narrowing regressions fail this test before merging.
* New `test_dialog_is_open_waits_for_late_mounting_dialog` — asserts the helper spends its budget on `wait_for(visible)` rather than reading `count()` eagerly.
* Existing `test_submit_invite_dialog_handles_two_button_gating_dialog` and the rest of `TestConnectWithPerson` continue to pass unchanged.
* Full suite: **510 passed** (was 508 before this commit).

## Constraint notes

I deliberately did **not** rip out the `a48fccc` button-count change. It is load-bearing once the dialog is detected — without it, the secondary-button click on the new 2-button gating layout would skip and the textarea would never mount. The right way to read this PR is "fix the upstream gate so the existing fix can actually run."

I also did not add a click-based fallback (profile → Connect button → modal). The deeplink approach works once the selector and timing are correct, and adding a fallback path would double the surface area for this tool. If the deeplink ever genuinely stops auto-mounting a modal (i.e. LinkedIn deprecates it), a separate PR can add the click-based path; the current code already has all the helpers (`_open_more_menu`, `click_button_by_text`, etc.) needed for that.

### DIFF
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -1559,15 +1559,25 @@ class LinkedInExtractor:
         if note:
             textarea_count = await self._page.locator(_DIALOG_TEXTAREA_SELECTOR).count()
             if textarea_count == 0:
-                # Reveal the note textarea via the secondary action. The note
-                # layout exposes three buttons (dismiss, secondary, primary);
-                # require all three before clicking the second-to-last so we
-                # never click dismiss on a no-note layout.
+                # Reveal the note textarea via the secondary action.
+                # Two layouts are now in the wild and both place "Add a
+                # note" at index ``btn_count - 2``:
+                #   * Legacy invite dialog (3 buttons): dismiss, secondary
+                #     "Add a note", primary "Send" -> nth(1) is secondary.
+                #   * "Add a note to your invitation?" gating dialog (2
+                #     buttons, rolled out 2026-05): "Add a note",
+                #     "Send without a note" -> nth(0) is the only path
+                #     that mounts the textarea. See issue #455.
+                # If LinkedIn ever serves a 2-button dismiss/primary
+                # no-note layout, the click below misroutes to dismiss;
+                # the textarea-presence recheck via _fill_dialog_textarea
+                # then fails and the caller returns connect_unavailable
+                # without sending — the same outcome as today.
                 buttons = self._page.locator(
                     f"{_DIALOG_SELECTOR} button, {_DIALOG_SELECTOR} [role='button']"
                 )
                 btn_count = await buttons.count()
-                if btn_count >= 3:
+                if btn_count >= 2:
                     await buttons.nth(btn_count - 2).click()
                     try:
                         await self._page.wait_for_selector(
--- a/tests/test_scraping.py
+++ b/tests/test_scraping.py
@@ -1393,6 +1393,78 @@ class TestConnectWithPerson:
 
         assert result["status"] == "unavailable"
 
+    async def test_submit_invite_dialog_handles_two_button_gating_dialog(
+        self, mock_page
+    ):
+        """Two-button "Add a note to your invitation?" gating dialog (issue
+        #455): nth(0) is "Add a note", nth(1) is "Send without a note".
+
+        Asserts the secondary-button click that reveals the textarea fires
+        even with btn_count == 2 (legacy guard required >= 3 and skipped
+        the click, leaving the textarea unmounted)."""
+        extractor = LinkedInExtractor(mock_page)
+
+        # Track each button click so we can assert the "Add a note" path
+        # was taken to reveal the textarea.
+        clicks: list[int] = []
+
+        textarea_visible = {"value": False}
+
+        # Two button locators inside the gating dialog: nth(0) "Add a
+        # note" reveals the textarea, nth(1) "Send without a note".
+        button_locators = [MagicMock(), MagicMock()]
+        for idx, btn in enumerate(button_locators):
+
+            def make_click(i: int):
+                async def _click(*args, **kwargs):
+                    clicks.append(i)
+                    if i == 0:
+                        textarea_visible["value"] = True
+                    return None
+
+                return _click
+
+            btn.click = AsyncMock(side_effect=make_click(idx))
+            btn.focus = AsyncMock()
+
+        button_collection = MagicMock()
+        button_collection.count = AsyncMock(return_value=2)
+        button_collection.nth = MagicMock(side_effect=lambda i: button_locators[i])
+
+        textarea_locator = MagicMock()
+        textarea_locator.count = AsyncMock(
+            side_effect=lambda: 1 if textarea_visible["value"] else 0
+        )
+        textarea_locator.first = textarea_locator
+        textarea_locator.fill = AsyncMock()
+
+        # Route page.locator() calls by selector — buttons vs textarea —
+        # so the gating dialog's button collection is distinguishable
+        # from the textarea probe.
+        def locator_router(selector: str):
+            if "textarea" in selector:
+                return textarea_locator
+            return button_collection
+
+        mock_page.locator = MagicMock(side_effect=locator_router)
+        mock_page.wait_for_selector = AsyncMock()
+        mock_page.keyboard = MagicMock()
+        mock_page.keyboard.press = AsyncMock()
+
+        with patch.object(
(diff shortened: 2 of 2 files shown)


stickerdaniel:
Thanks for the thorough writeup. Verified against my account: LinkedIn serves me a 3-button variant of the same gating dialog (`[Dismiss, Add a note, Send without a note]`), so the `>= 3` guard passed by coincidence. Lowering to `>= 2` makes both layouts converge on the correct `nth(btn_count - 2)`.

Cherry-picked your new test onto `main` to confirm: fails there, passes here. Full suite 174/174.

Issue #454, opened 2026-05-18
[BUG] connect_with_person returns connect_unavailable ("did not expose a usable Connect action") on profiles where Connect is clearly the primary CTA — v4.13.0


PasSivam:
# [BUG] `connect_with_person` returns `connect_unavailable` ("did not expose a usable Connect action") on profiles where Connect is clearly the primary CTA — v4.13.0

## Summary

On `linkedin-scraper-mcp` **v4.13.0**, `connect_with_person` consistently returns:

```json
{
  "status": "connect_unavailable",
  "message": "LinkedIn did not expose a usable Connect action for this profile.",
  "note_sent": false
}
```

on profiles whose **own profile DOM (returned by the same tool call) explicitly contains a `Connect` button as the primary CTA**, right alongside `Save in Sales Navigator` and `More`. The tool is seeing Connect but refusing to act on it.

The error string `"did not expose a usable Connect action"` does not appear in any open or closed issue today, so this looks like an unreported variant — possibly distinct from #407 (which is about restricted-invite profiles where the *dialog* opens but the *Send* button is disabled, with error wording `"did not open a usable invite dialog"`).

## Environment

- **Package:** `linkedin-scraper-mcp` **v4.13.0** (installed via `uvx`, latest on PyPI as of 2026-05-18)
- **Client:** Claude Code (MCP stdio transport)
- **Platform:** macOS 25.4.0 (darwin, aarch64)
- **uvx:** 0.11.14
- **Cache state:** Bug persists after `uvx --refresh linkedin-scraper-mcp` (full reinstall from PyPI)
- **Session:** `--status` reports `Session is valid` (persistent profile at `~/.linkedin-mcp/profile`)
- **Invocation:** stdio, no custom args (`uvx linkedin-scraper-mcp`)

## Reproduction

Three profiles attempted back-to-back, all returned the same status. No interleaving with `send_message` (rules out #432). No `note=` parameter (rules out the noted-invite path).

| # | Profile (anonymized) | Followers | Connection degree | Profile DOM shows `Connect`? | Result |
|---|---|---|---|---|---|
| 1 | Profile A — high-follower account in "creator" or follow-first display mode | ~2.3K | 2nd | **No** — DOM shows `Follow / Save in Sales Navigator / View my services / More` | `connect_unavailable` (plausibly correct — Connect is genuinely not exposed) |
[...]

PR #458, opened 2026-05-21
fix(connect): detect alertdialog/aria-modal and wait for late-mounting invite modal

### LINKED ISSUE #432: [BUG] connect_with_person with note: dialog locator collides with message composer overlay (Emoji button)
## Setup

- Client: Claude Code CLI calling `mcp__linkedin__connect_with_person`
- OS: linux-amd64-host
- Trace artifacts dir: `~/.linkedin-mcp/trace-runs/run-zobpr291/`

## What Happened

`connect_with_person(linkedin_username=..., note="...")` raises an uncaught Playwright timeout. The trace shows the dialog locator resolves to the **Emoji Keyboard button** of the **message composer overlay**, not the invite dialog. Click times out because that emoji button is offscreen.

Reproduction:
- Same call WITHOUT `note=` → succeeds (`status: connected`, dialog opens, request sent)
- Same call WITH `note=` (any string, ≤300 chars) → fails with generic `Error calling tool 'connect_with_person'`

The asymmetry is deterministic across multiple recipients tested today: kirtichablani, babitha-s-508590b3, palaniappan-lakshmanan-6a153a20.

## Suspected Root Cause

From `server.log` excerpt (run-zobpr291):

```
Unexpected error in connect_with_person: Locator.click: Timeout 5000ms exceeded.
Call log:
  - waiting for locator("dialog[open], [role=\"dialog\"] button, dialog[open], [role=\"dialog\"] [role='button']").nth(22)
    - locator resolved to <button id="ember125" type="button" aria-expanded="false" 
        title="Open Emoji Keyboard" aria-label="Open Emoji Keyboard" 
        data-reaction-summary__emoji-hoverable="" 
        aria-controls="artdeco-hoverable-msg_overlay_emoji__emoji-hoverable__content" 
        class="msg-form__footer-action emoji-hoverable-trigger ...">
    - element is outside of the viewport
```

The selector `dialog[open], [role="dialog"] button` matches BOTH the connect-invite dialog AND the LinkedIn message composer overlay (which is also rendered as `[role="dialog"]`). When a previous `send_message` call has left the message composer mounted on the page, the `.nth(22)` index lands on a button inside the composer (here: Open Emoji Keyboard), not the invite dialog's Send button.

This explains why the bug only appears when `note=` is passed: the noteless path uses a different control flow that doesn't depend on `.nth(22)`.

## Steps to Reproduce

1. From a fresh session, call `send_message(linkedin_username="x", message="hi", confirm_send=True)` for any 1st-degree contact (this mounts the message composer overlay, even after dialog closes).
2. Then call `connect_with_person(linkedin_username="y", note="hi from me")` for any 2nd-degree contact whose Connect button is visible.
3. Observe `Error calling tool 'connect_with_person'`.
4. Repeat without `note=` → succeeds.

## Suggested Fix

Scope the dialog locator to exclude the message composer overlay — e.g. `[role="dialog"]:not([id^="artdeco-hoverable-msg_overlay"]) button`, or scope by parent test-id that's specific to invite dialogs. The current `.nth(N)` indexing into a multi-dialog DOM is brittle.

## Logs

Full trace: `~/.linkedin-mcp/trace-runs/run-zobpr291/server.log` (can attach if useful).

ivangotti:
Follow-up to #456 (merged). The 2-button-gating handling shipped in #456 is correct *for the case where the dialog actually opens*, but on real accounts post-deploy `connect_with_person` was still returning `connect_unavailable` for every with-note attempt — because `_dialog_is_open` was failing one layer upstream and the secondary-button branch never ran.

This PR fixes the upstream gate. The 2-button-gating handling from #456 is preserved verbatim and is now reachable.

Re #455 (post-#456 follow-up). Distinct from #454 (different failure stage — that one bails before reaching the deeplink gate) and #432 (composer-overlay collision — orthogonal).

## Real root cause

Two issues in `_dialog_is_open` / `_DIALOG_SELECTOR`, both visible in trace `run-8fic76qw` from a failing 8/8 production run on 2026-05-20 (post-#456 deploy):

1. **Selector was too narrow.** The "Add a note to your invitation?" gating dialog is an `artdeco-modal` whose root carries `role="alertdialog"` and/or `aria-modal="true"`, neither of which matched `dialog[open], [role="dialog"]`. Trace screenshots show the dialog visibly mounted at the exact DOM moment `_dialog_is_open` was returning False — the helper just couldn't see it.

2. **Race against modal mount.** `_dialog_is_open` short-circuited on `count() == 0` before spending its timeout budget. After `goto(/preload/custom-invite/?vanityName=)` resolves on `domcontentloaded`, the modal mounts a few hundred ms later, so the count check fired against an empty DOM and returned False immediately. `wait_for(state="visible")` is the right primitive — it polls until the element exists *and* is visible.

The two effects compound. Even on accounts where LinkedIn picks `role="dialog"` and the selector would have matched, the eager `count()` made the helper racy. Even on a slow account where waiting longer would have helped, the wrong selector would still miss the alertdialog mount.

## Evidence

From the trace artifact `~/.linkedin-mcp/trace-runs/run-8fic76qw/` (real failing run, 2026-05-20, account-wide reproduction matching #455's reporter):

* Step 5: `goto preload/custom-invite/?vanityName=meenal-ganesh-b1710321`. Screenshot `005-extractor-after-goto.png` shows the gating dialog **fully mounted and visible** (X close + "Add a note" + "Send without a note"). Body marker: `"Dialog content start. Add a note to your invitation? Personalize your invitation to Meenal Ganesh by adding a note."` — verbatim DOM text.
* The dialog is structurally identical across Brian W., Igor Dimoski, Sandro Valentini (screenshots 013, 017, 021).
* For Greg Sandzimier (run-3mu94g3t, 2026-05-19, *before* the gating dialog rolled out broadly to this account), the trace shows a verification scrape immediately after the deeplink — i.e. `submitted=True`, dialog detected, "Send without a note" clicked, profile transitioned to `"More Pending Message"`. Invite landed.
* For Brian / Igor / Sandro on 2026-05-20, there is **no verification scrape after the deeplink** — the trace jumps straight to the next person's profile. `submitted=False` → `connect_unavailable`. Same dialog DOM, different `_dialog_is_open` result. Only the LinkedIn-side rollout state differed.

## The fix

```diff
 _DIALOG_SELECTOR = (
+    'dialog[open], '
+    '[role="dialog"], '
+    '[role="alertdialog"], '
+    '[aria-modal="true"]'
 )

 _DIALOG_TEXTAREA_SELECTOR = (
+    '[role="dialog"] textarea, '
+    '[role="alertdialog"] textarea, '
+    '[aria-modal="true"] textarea, '
+    'dialog textarea'
 )

 async def _dialog_is_open(self, *, timeout: int = 1000) -> bool:
-    locator = self._page.locator(_DIALOG_SELECTOR)
+    locator = self._page.locator(_DIALOG_SELECTOR).first
     try:
-        if await locator.count() == 0:
-            return False
-        await locator.first.wait_for(state="visible", timeout=timeout)
+        await locator.wait_for(state="visible", timeout=timeout)
         return True
     except Exception:
         return False
```

Per-call sites: every existing reference to `_DIALOG_SELECTOR` and `_DIALOG_TEXTAREA_SELECTOR` (button collection in `_submit_invite_dialog`, textarea fill in `_fill_dialog_textarea`, dismiss-on-hidden) gets the broadened scope automatically.

The 2-button-gating handling from #456 is preserved — with the dialog now correctly detected, the existing `btn_count >= 2` / `nth(btn_count - 2)` branch fires and reveals the textarea on the new gating layout, while the no-note path still hits "Send without a note" via `_click_dialog_primary_button`.

## Tests

* New `test_dialog_selector_matches_alertdialog_and_aria_modal` — guards every required ARIA pattern in `_DIALOG_SELECTOR` and `_DIALOG_TEXTAREA_SELECTOR`. Future narrowing regressions fail this test before merging.
* New `test_dialog_is_open_waits_for_late_mounting_dialog` — asserts the helper spends its budget on `wait_for(visible)` rather than reading `count()` eagerly.
* Existing `test_submit_invite_dialog_handles_two_button_gating_dialog` (from #456) and the rest of `TestConnectWithPerson` continue to pass unchanged.
* Full suite: **510 passed** (was 508 before this branch).

## Verification confidence

This is best-effort code reading + DOM trace evidence — I did not drive a real browser to verify the new selector against the live LinkedIn DOM. What I do know:

* The gating dialog is visible in the real trace screenshots.
* The body's accessible text (`"Dialog content start. ..."`) is fetched via `document.body.innerText`, so the dialog is in the DOM.
* `_dialog_is_open` was returning False against that DOM; only an out-of-scope selector or a race against mount can produce that.
* The replacement selector covers the three modal-root ARIA patterns LinkedIn ships across artdeco-modal versions, and the `wait_for(visible)` pattern is what the rest of the codebase already uses for modal detection (cf. `extract_overlay_page` and `handle_modal_close`).

If LinkedIn ever serves a modal with neither `role` attribute nor `aria-modal`, this fix won't catch it — but I've not seen evidence of that variant in any trace, and adding `.artdeco-modal__content` would risk matching non-modal artdeco UI elsewhere on the page.

## What this PR does *not* do

* Does **not** revert #456's `btn_count >= 2` change — it's load-bearing once the dialog is detected.
* Does **not** add a click-based fallback (profile → Connect button → modal). The deeplink approach works once the selector and timing are correct, and adding a fallback path would double the surface area for this tool. If the deeplink ever genuinely stops auto-mounting a modal (i.e. LinkedIn deprecates it), a separate PR can add the click-based path; the codebase already has `_open_more_menu` and `click_button_by_text` helpers for that.
* Does **not** fix #432 (composer-overlay button-index collision) or #454 (no-note `"did not expose a usable Connect action"` on profiles with visible Connect). Those are separate failure stages with different signatures.

### DIFF
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -768,7 +768,7 @@ class LinkedInExtractor:
         times out, so callers can fall back to a keyboard submit.
         """
         buttons = self._page.locator(
-            f"{_DIALOG_SELECTOR} button, {_DIALOG_SELECTOR} [role='button']"
+            f":is({_DIALOG_SELECTOR}) button, :is({_DIALOG_SELECTOR}) [role='button']"
         )
         count = await buttons.count()
         if count == 0:
@@ -1591,7 +1591,7 @@ class LinkedInExtractor:
                 # then fails and the caller returns connect_unavailable
                 # without sending — the same outcome as today.
                 buttons = self._page.locator(
-                    f"{_DIALOG_SELECTOR} button, {_DIALOG_SELECTOR} [role='button']"
+                    f":is({_DIALOG_SELECTOR}) button, :is({_DIALOG_SELECTOR}) [role='button']"
                 )
                 btn_count = await buttons.count()
                 if btn_count >= 2:
@@ -1616,7 +1616,7 @@ class LinkedInExtractor:
             # Enter targets it instead of a focused textarea (where Enter
             # would just insert a newline).
             buttons = self._page.locator(
-                f"{_DIALOG_SELECTOR} button, {_DIALOG_SELECTOR} [role='button']"
+                f":is({_DIALOG_SELECTOR}) button, :is({_DIALOG_SELECTOR}) [role='button']"
             )
             btn_count = await buttons.count()
             if btn_count > 0:
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -93,8 +93,18 @@ _SORT_BY_MAP = {"date": "DD", "relevance": "R"}
 # LinkedIn accepts "F" (1st-degree), "S" (2nd-degree), "O" (3rd-degree and beyond).
 _NETWORK_TOKENS = ("F", "S", "O")
 
-_DIALOG_SELECTOR = 'dialog[open], [role="dialog"]'
-_DIALOG_TEXTAREA_SELECTOR = '[role="dialog"] textarea, dialog textarea'
+_DIALOG_SELECTOR = (
+    'dialog[open], '
+    '[role="dialog"], '
+    '[role="alertdialog"], '
+    '[aria-modal="true"]'
+)
+_DIALOG_TEXTAREA_SELECTOR = (
+    '[role="dialog"] textarea, '
+    '[role="alertdialog"] textarea, '
+    '[aria-modal="true"] textarea, '
+    'dialog textarea'
+)
 
 _MESSAGING_COMPOSE_LINK_SELECTOR = 'main a[href*="/messaging/compose/"]'
 _MESSAGING_COMPOSE_SELECTOR = (
@@ -733,12 +743,19 @@ class LinkedInExtractor:
             return False
 
     async def _dialog_is_open(self, *, timeout: int = 1000) -> bool:
-        """Return whether a dialog is currently open (structural check)."""
-        locator = self._page.locator(_DIALOG_SELECTOR)
+        """Return whether a dialog is currently open (structural check).
+
+        Uses ``wait_for(state="visible", timeout=timeout)`` directly so the
+        full timeout budget is spent waiting for the dialog to mount —
+        important after navigations that auto-open a modal (e.g. the
+        ``/preload/custom-invite/`` deeplink), where the modal mounts a
+        few hundred ms after ``goto`` resolves on ``domcontentloaded``.
+        The previous early ``count() == 0`` short-circuit returned False
+        before the modal had a chance to render.
+        """
+        locator = self._page.locator(_DIALOG_SELECTOR).first
         try:
-            if await locator.count() == 0:
-                return False
-            await locator.first.wait_for(state="visible", timeout=timeout)
+            await locator.wait_for(state="visible", timeout=timeout)
             return True
         except Exception:
             return False
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -93,18 +93,8 @@ _SORT_BY_MAP = {"date": "DD", "relevance": "R"}
 # LinkedIn accepts "F" (1st-degree), "S" (2nd-degree), "O" (3rd-degree and beyond).
 _NETWORK_TOKENS = ("F", "S", "O")
 
-_DIALOG_SELECTOR = (
-    'dialog[open], '
-    '[role="dialog"], '
-    '[role="alertdialog"], '
-    '[aria-modal="true"]'
-)
-_DIALOG_TEXTAREA_SELECTOR = (
-    '[role="dialog"] textarea, '
-    '[role="alertdialog"] textarea, '
-    '[aria-modal="true"] textarea, '
-    'dialog textarea'
-)
+_DIALOG_SELECTOR = 'dialog[open], [role="dialog"]'
+_DIALOG_TEXTAREA_SELECTOR = '[role="dialog"] textarea, dialog textarea'
 
 _MESSAGING_COMPOSE_LINK_SELECTOR = 'main a[href*="/messaging/compose/"]'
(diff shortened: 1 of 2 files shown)


ivangotti:
Greptile's review caught a real bug — confirmed in production.

After the dialog-selector broadening on this PR, my own invite attempts against `alertdialog`-rooted profiles still returned `connect_unavailable`. Greptile's diagnosis is exactly right: the descendant button selectors built from `{_DIALOG_SELECTOR} button` only apply the descendant combinator to the LAST arm of the comma-separated selector list. The other arms match dialog roots instead of buttons, so the click goes to the modal root and "Send" never fires.

Pushed [50a695b](https://github.com/ivangotti/linkedin-mcp-server/commit/50a695b) wrapping `_DIALOG_SELECTOR` in `:is()` at all 3 descendant-button locator sites. Matches Greptile's suggested fix verbatim.

Will verify against my account and report back.

ivangotti:
Found the actual root cause via live-DOM probe.

**What was happening:** When the free-note monthly quota is exhausted, LinkedIn replaces the 'Add a note?' gating dialog with a 'Send unlimited personalized invites with Premium — You're out of free custom notes' upsell modal. The code waited for the textarea (which never mounts), then dismissed and returned `connect_unavailable` — discarding the connection request entirely.

The Greptile-flagged `:is()` and selector-broadening issues both turned out to be diagnosis red herrings. Reverted those (`11540ba`) and replaced with a targeted upsell-detection + no-note fallback (`462ec7b`).

Probe-validated end-to-end against live LinkedIn:
- 'Add a note' click → upsell appears (URL unchanged at `/preload/custom-invite/`)
- Detect via `:has-text("free custom notes")` / "personalized invites with Premium"
- Dismiss upsell → re-navigate to the same invite URL (re-triggers the original dialog)
- Click 'Send without a note' → invite sent → profile transitions to Pending

Caller still sees `note_sent=False` so the dropped personalization is visible.

Verified against `babermehkeri` and `javargas209` — both connections sent successfully via this path.

PR #465, opened 2026-05-26
fix(messaging): dismiss stale message overlay before send_message

### LINKED ISSUE #433: [BUG] send_message: composer_unavailable on second+ call in same session, even for 1st-degree contacts
## Setup

- Client: Claude Code CLI calling `mcp__linkedin__send_message`
- OS: linux-amd64-host

## What Happened

`send_message` returns `{"status": "composer_unavailable", "message": "LinkedIn did not expose a usable message composer.", "recipient_selected": false, "sent": false}` on the **second and subsequent** calls in the same session, even when:

- The recipient is a 1st-degree connection
- The recipient profile is reachable (verified via `get_person_profile` returning the `Message` button)
- The previous `send_message` call to the same recipient succeeded with `status: sent`

## Reproduction (deterministic, observed today across multiple sessions)

```
Call 1: send_message(linkedin_username="palaniappan-lakshmanan-6a153a20", message="Hi P — ...")
        → {"status": "sent", "message": "Message sent.", "sent": true}   ✅

Call 2: send_message(linkedin_username="palaniappan-lakshmanan-6a153a20", message="<para 2 of multi-paragraph>")
        → {"status": "composer_unavailable", "sent": false}   ❌

Call 3: send_message(linkedin_username="palaniappan-lakshmanan-6a153a20", message="<para 2 retry>")
        → {"status": "composer_unavailable", "sent": false}   ❌
```

This breaks the documented workaround for `\n\n` truncation (split message into per-paragraph send_message calls in sequence). After the first paragraph lands, no further paragraphs can be sent in the same session.

## Why This Matters

LinkedIn's actual messaging API truncates messages at the first blank line (`\n\n`). The standard workaround is to send each paragraph as a separate `send_message` call. With this bug, only the first paragraph ever lands — the closing CTA + signoff get silently dropped.

For the user, this means cold outreach delivers as: `"Hi <name> — [opening line about the role]"` with no body, no ask, no signoff. Embarrassing in production use.

## Related (closed)

- #344 (closed) — `send_message returns composer_unavailable consistently (headless mode)` — symptoms appear similar but #344 was a cold-start issue. This new occurrence happens *after* a successful first send, suggesting a state-management regression.
- #296 (closed) — `hardcoded 5s timeout causes composer_unavailable` — possibly related if the second-call composer-mount is racing, but the first call succeeds in <2s, so timeout doesn't seem to be the constraint.

## Suggested Investigation

The composer overlay mounted by Call 1 may be in a stale state when Call 2 runs. The MCP needs to either: (a) close-and-remount the composer between sends, (b) reuse the open composer if it's still pointed at the same recipient, or (c) navigate away and back to the profile to force a fresh mount.

## Logs

No specific trace artifact was generated (this is a returned-error response, not an uncaught exception, so the auto-trace didn't fire). Happy to repro on demand.

devag7:
Fixes #433. Calls _dismiss_message_ui at the beginning of send_message to prevent stale overlays from causing composer_unavailable.

### DIFF
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -3167,6 +3167,10 @@ class LinkedInExtractor:
             profile_urn: Optional profile URN (e.g. ACoAAB...) to construct the
                 compose URL directly, bypassing the Message-button lookup.
         """
+        # Clear any floating message overlay from previous tools/calls
+        # that might intercept the compose box or block page interactions.
+        await self._dismiss_message_ui()
+
         profile_url = f"https://www.linkedin.com/in/{linkedin_username}/"
         await self._navigate_to_page(profile_url)
         await detect_rate_limit(self._page)
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -3167,10 +3167,6 @@ class LinkedInExtractor:
             profile_urn: Optional profile URN (e.g. ACoAAB...) to construct the
                 compose URL directly, bypassing the Message-button lookup.
         """
-        # Clear any floating message overlay from previous tools/calls
-        # that might intercept the compose box or block page interactions.
-        await self._dismiss_message_ui()
-
         profile_url = f"https://www.linkedin.com/in/{linkedin_username}/"
         await self._navigate_to_page(profile_url)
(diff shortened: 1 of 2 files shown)


devag7:
Superseded by #482, which consolidates this fix with all reviewer feedback addressed.

PR #468, opened 2026-05-26
fix(docker): set UV_PYTHON_INSTALL_DIR to avoid permission denied

devag7:
Fixes #321. Sets UV_PYTHON_INSTALL_DIR during docker build to prevent uv from using root-owned symlinks, allowing pwuser to run python.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,14 @@
+ARG UV_PYTHON_INSTALL_DIR=/opt/uv-python
+
 # -- Stage 1: Build virtual environment --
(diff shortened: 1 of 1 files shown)


stickerdaniel:
Hey, thanks for the PR! Two things though: #321 is already fixed on `main` in `79c193c`, which pins the base image to python 3.13 so uv uses the system interpreter in place and never pulls a managed one into `/root`. And with that pinned base, uv doesn't create `/opt/uv-python` at all, so `COPY --from=builder /opt/uv-python` actually fails the build with `"/opt/uv-python": not found` (the PR checks stay green because only the release workflow builds the image). Going to close this one, but thanks for digging into the Docker setup!

devag7:
Superseded by #482, which includes the Dockerfile improvement.

PR #475, opened 2026-05-30
fix(messaging): resolve get_conversation via inbox

### LINKED ISSUE #434: [BUG] get_conversation cannot find thread that was just created via send_message (2nd-degree recipient)
## Setup

- Client: Claude Code CLI calling `mcp__linkedin__get_conversation`
- OS: linux-amd64-host
- Trace artifacts dir: `~/.linkedin-mcp/trace-runs/run-zobpr291/` (auto-generated by the MCP — includes `screens/`, `trace.jsonl`, `server.log`)

## What Happened

`get_conversation(linkedin_username="palaniappan-lakshmanan-6a153a20")` returns:

```
LinkedInScraperException: Could not find a conversation for palaniappan-lakshmanan-6a153a20.
```

…immediately after a successful `send_message` to the same recipient that returned `{"status": "sent", "message": "Message sent.", "sent": true}`.

The thread *is* visible in the LinkedIn UI when opened in a browser — the message landed in the recipient's inbox per recipient confirmation. So the conversation exists; the MCP enumeration just can't find it.

## Reproduction

```
Call 1: send_message(linkedin_username="palaniappan-lakshmanan-6a153a20", message="Hi P — ...")
        → {"status": "sent", "sent": true}   ✅ (verified by recipient: message received)

Call 2: get_conversation(linkedin_username="palaniappan-lakshmanan-6a153a20")
        → LinkedInScraperException: Could not find a conversation for palaniappan-lakshmanan-6a153a20.   ❌
```

## Why This Matters

The documented workflow for sending multi-paragraph messages is to split at `\n\n` and send each paragraph sequentially via `send_message`, then verify with `get_conversation`. The verification step is essential because `send_message` itself is unreliable (see #433 for `composer_unavailable` on second call). When `get_conversation` also fails to find the just-created thread, there's no programmatic way to confirm what landed.

For the calling agent / human-in-the-loop, this means the only reliable verification is the user manually opening LinkedIn — defeating the automation.

## Hypotheses

1. **Recency lag** — LinkedIn may not surface a brand-new thread in the inbox enumeration immediately after the first message. The MCP enumerates by clicking through inbox rows; if the new thread hasn't propagated to the inbox-list rendering yet, no row matches.
2. **InMail vs Direct Message routing** — for 2nd-degree recipients, the message may be routed to the InMail tab or Connection-request tab rather than Focused. The enumeration may only check Focused.
3. **Display name mismatch** — the docstring notes thread resolution searches by participant *display name*. If the participant's name was rendered with a He/Him suffix or similar in the inbox row, the substring match may fail.

## Related (closed)

- #307 (closed) — `get_conversation fails when user has multiple threads with the same person`. Different scenario (multi-thread); mine is a single fresh thread that the enumeration can't find at all.

## Steps to Reproduce

1. Send a message to a 2nd-degree connection via `send_message`.
2. Immediately call `get_conversation` for the same `linkedin_username`.
3. Observe `LinkedInScraperException: Could not find a conversation`.
4. Manually open LinkedIn UI — the thread is visible.

## Logs / Trace Artifacts

Auto-generated artifacts in `~/.linkedin-mcp/trace-runs/run-zobpr291/`:
- `20260506T160712Z-get-conversation.md` — issue template
- `screens/` — captured browser screenshots
- `trace.jsonl` — full trace
- `server.log` — server-side log

Happy to attach any of these as a follow-up comment if useful.

stickerdaniel:
Closes #434

`get_conversation(linkedin_username=...)` raised "Could not find a conversation" even when the thread existed and was fully readable by `thread_id`. This broke the verify-after-send workflow (send_message then get_conversation), leaving no programmatic way to confirm what landed.

The username-to-thread resolver navigated to `/messaging/?searchTerm={display_name}` and enumerated the search results. LinkedIn's messaging search returns "We didn't find anything" for the exact display name even when the identical thread is the top row of the plain inbox, so resolution found nothing. The issue's hypotheses (recency lag, InMail routing, name mismatch) don't hold: the thread was hours old, 1st-degree, and the inbox row's aria-label matched the profile name exactly.

The resolver now enumerates the plain inbox (`/messaging/`) the same way `get_inbox` already does, and passes a `name_filter` to the row enumerator so only the row matching the participant is clicked. Clicking a row can mark it read, so the filter keeps that side effect scoped to the requested participant instead of touching every recent thread. The exact-name match and multi-thread `index` selection are unchanged, and `get_inbox`/`search_conversations` are unaffected (the new parameter defaults to off).

`uv run pytest` passes (510 tests, including updated and new coverage for the inbox resolver). Verified live against a real account: `get_conversation` by username now returns the thread it previously couldn't find, and `get_inbox` still enumerates unchanged.

### DIFF
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -93,6 +93,12 @@ _SORT_BY_MAP = {"date": "DD", "relevance": "R"}
 # LinkedIn accepts "F" (1st-degree), "S" (2nd-degree), "O" (3rd-degree and beyond).
 _NETWORK_TOKENS = ("F", "S", "O")
 
+# How many inbox rows to iterate when resolving a conversation by participant
+# name. Only the matching row is clicked (see _extract_conversation_thread_refs
+# name_filter), so iterating a generous count is cheap; recent threads (the
+# common get_conversation-by-username case) sit near the top of the inbox.
+_INBOX_RESOLVE_LIMIT = 50
+
 _DIALOG_SELECTOR = 'dialog[open], [role="dialog"]'
 _DIALOG_TEXTAREA_SELECTOR = '[role="dialog"] textarea, dialog textarea'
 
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#475** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/475?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/475?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
Thanks! Triaged all three:

1. Inbox scroll skips recent rows: false positive. Verified live, the inbox appends on scroll (10 to 17 rows) rather than virtualizing out the top, so the recent thread stays at index 0 and resolves fine after the scroll.

2. Limit applies before filtering: fixed in 0f47a31. The resolver now passes `limit=None`, so every rendered row is scanned and the name filter keeps clicking scoped to the match, removing the row-50 cliff.

3. Search fallback removed: re-added in 0f47a31 as a last resort. The inbox scan runs first (reliable for recent threads); if it finds nothing, resolution falls back to the messaging search for threads buried below the inbox window.

stickerdaniel:
On the duplicate-thread `index` path: documented rather than changed (c41233d). It needs a buried duplicate thread plus `index>=1` together, and InMail threads do appear in the plain inbox so they are covered by the scan. The `index` contract already routes multi-thread disambiguation to `search_conversations` + `thread_id`, now called out in the resolver docstring. Keeping inbox-first with search-only-as-fallback so the common path stays off the unreliable messaging search.

PR #478, opened 2026-05-30
fix(docker): add tini init to reap zombie chromium processes

### LINKED ISSUE #477: Docker container leaks zombie chrome-headless processes (no init/reaper)
### Description

When the server runs in Docker as a long-lived process, exited Chromium subprocesses accumulate as zombie (`<defunct>`) processes and never get reaped. Over a multi-day-lived container these pile up.

### Steps to reproduce

1. Run the published image as a long-running service, e.g. streamable-http:
   ```bash
   docker run -d --name linkedin-mcp \
     -v ~/.linkedin-mcp:/home/pwuser/.linkedin-mcp \
     stickerdaniel/linkedin-mcp-server:latest \
     --transport streamable-http --host 0.0.0.0 --port 8080 --path /mcp
   ```
2. Issue a few scraping calls so the bridge/scrape flow launches Chromium.
3. Inspect the container's process tree:
   ```bash
   docker top linkedin-mcp
   # or, on the host:
   ps -ef | grep defunct
   ```

### Observed

Several `[chrome-headless] <defunct>` processes parented to the Python PID, e.g.:

```
python(1)-+-chrome-headless(...)   <- Z (zombie)
          |-chrome-headless(...)   <- Z (zombie)
          ...
```

In my case 4 zombies appeared within ~20s of the first request and sat there for 9 days.

### Root cause

`Dockerfile` sets `ENTRYPOINT ["python", "-m", "linkedin_mcp_server"]`, so **Python runs as PID 1**. When Patchright/Chromium helper processes exit, the kernel reparents them to PID 1, but Python never calls `waitpid()` (no `SIGCHLD` handler / asyncio child watcher), so they remain as zombies forever. On a normal host this is invisible because real `init` reaps orphans; inside a container with no init, only PID 1 can reap and Python doesn't.

### Proposed fix

Add a minimal init (`tini`) as the entrypoint so PID 1 reaps orphaned subprocesses:

```dockerfile
RUN apt-get update && apt-get install -y --no-install-recommends tini && rm -rf /var/lib/apt/lists/*
ENTRYPOINT ["tini", "--", "python", "-m", "linkedin_mcp_server"]
```

This fixes the published image for all consumers, including plain `docker run` users (i.e. without relying on `docker run --init` / compose `init: true`).

### Environment

- Image: `stickerdaniel/linkedin-mcp-server:latest`
- Transport: streamable-http
- Host: Linux

manoj1919:
## Summary

The container runs `python` as PID 1 (`ENTRYPOINT ["python", "-m", "linkedin_mcp_server"]`). When Patchright/Chromium helper processes exit, the kernel reparents them to PID 1, but Python never `waitpid()`s them, so they accumulate as zombie (`<defunct>`) `chrome-headless` processes for the life of the container.

This adds `tini` as a minimal init so PID 1 reaps orphaned subprocesses.

```dockerfile
RUN apt-get update && apt-get install -y --no-install-recommends tini && rm -rf /var/lib/apt/lists/*
ENTRYPOINT ["tini", "--", "python", "-m", "linkedin_mcp_server"]
```

Fixing it in the image (rather than telling users to pass `--init` / compose `init: true`) means plain `docker run` consumers are covered too.

## Verification

Built the image and confirmed PID 1 inside the container is now an init wrapping Python, with no zombies after exercising the Chromium-launching bridge flow:

```
PID 1: /sbin/...init -- python -m linkedin_mcp_server ...
process tree: init(1)---python(...)
zombies: none
```

Closes #477

## Synthetic prompt

> The Dockerized server leaks zombie `chrome-headless` processes because Python runs as PID 1 with no init to reap orphaned Chromium subprocesses. Add `tini` to the Dockerfile and make it the entrypoint (`tini -- python -m linkedin_mcp_server`) so PID 1 reaps zombies, fixing it at the image level so plain `docker run` users benefit without `--init`.

Generated with Claude Opus 4.8


stickerdaniel:
Hey, thanks for this! Built both release platforms and confirmed the reaping. Also fixes `docker stop` taking 10s: python as PID 1 never sees SIGTERM, so the container got SIGKILLed and the browser cleanup never ran.

Issue #389, opened 2026-04-23
[BUG] TypeError: _eval_type() got an unexpected keyword argument 'prefer_fwd_module'


Spenhouet:
### Setup

.mcpb bundle with Claude Desktop on Windows

### What Happened

I installed the bundle and when the MCP got enabled it reported 

<img width="1308" height="480" alt="Image" src="https://github.com/user-attachments/assets/f9d5d808-8bd2-41c5-8938-54c7447ee477" />

Restart of Claude Desktop or disabling and reenabling the MCP did not help.

### Steps to Reproduce

1. In Claude Connector Settings select "Add" -> "Desktop"
2. Select the .mcpb file of the 4.9.3 release
3. Wait for the install to finish
4. -> Error

### Logs

```shell
2026-04-23T15:13:03.137Z [LinkedIn MCP Server] [info] Message from client: {"method":"initialize","params":{"protocolVersion":"2025-11-25","capabilities":{"extensions":{"io.modelcontextprotocol/ui":{"mimeTypes":["text/html;profile=mcp-app"]}}},"clientInfo":{"name":"claude-ai","version":"0.1.0"}},"jsonrpc":"2.0","id":0} { metadata: undefined }
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\spe\AppData\Roaming\Claude\Claude Extensions\ant.dir.ant.0x8afe944.linkedin-mcp-server\linkedin_mcp_server\__main__.py", line 4, in <module>
    from linkedin_mcp_server.cli_main import main
  File "C:\Users\spe\AppData\Roaming\Claude\Claude Extensions\ant.dir.ant.0x8afe944.linkedin-mcp-server\linkedin_mcp_server\cli_main.py", line 10, in <module>
    from linkedin_mcp_server.bootstrap import (
    ...<2 lines>...
    )
  File "C:\Users\spe\AppData\Roaming\Claude\Claude Extensions\ant.dir.ant.0x8afe944.linkedin-mcp-server\linkedin_mcp_server\bootstrap.py", line 16, in <module>
    from fastmcp import Context
  File "C:\Users\spe\AppData\Roaming\Claude\Claude Extensions\ant.dir.ant.0x8afe944.linkedin-mcp-server\.venv\Lib\site-packages\fastmcp\__init__.py", line 8, in <module>
    from fastmcp.settings import Settings
[...]

PR #490, opened 2026-06-08
fix: relax .python-version to 3.13 minor instead of exact patch

ca-aford-gtm:
Fixes the `No interpreter found for Python 3.13.13` MCPB launch failure — the
bundled server dies on first launch when uv can't resolve that exact patch:

    error: No interpreter found for Python 3.13.13 in managed installations or search path

`.python-version` was pinned to the full triple `3.13.13` (40f8797, to dodge the
3.14.0rc1 cache trap from #389). But `uv run` reads `.python-version` and demands
that exact patch — if the user has no 3.13.13 and their uv's managed-download list
has no python-build-standalone build for that precise patch, resolution fails.

Pinning the minor `3.13` lets uv select any stable 3.13.x (downloading if needed),
while still avoiding the original rc trap: that only bit because `3.14` was
bleeding-edge with an rc cached. 3.13 is mature and uv won't pick prereleases for a
plain minor. requires-python (>=3.12,<3.15) already guarantees any 3.13.x works.

The Dockerfile keeps its 3.13.13 digest pin (real Docker Hub image, unaffected).


stickerdaniel:
Hey, thanks for the PR! The same change landed on main in ea01f52 shortly after you opened this, .python-version is now 3.13. Closing as superseded.

PR #492, opened 2026-06-08
fix(mcpb): loosen bundled Python pin so installs don't require an exact patch

### LINKED ISSUE #484: [BUG] Exact Patch Release Required 3.13
### Setup

macOS, Claude Desktop

### What Happened

AI debug: The underlying problem is the bundle author's: pinning an exact patch release that's two months old is brittle and will break for anyone whose resolver index lags. That's worth reporting to whoever publishes the bundle, and a fix on their end (loosening to >=3.13 or similar) is the real solution.

Image

### Steps to Reproduce

Install on system with newer version of python

### Logs

```shell

```

stickerdaniel:
Closes #484

The `.mcpb` bundle ships a `.python-version` file pinned to the exact patch `3.13.13`. When Claude Desktop launches the bundle with `uv run`, uv reads that file and demands exactly 3.13.13. On a machine whose uv has no 3.13.13 installed and does not list it in its python-build-standalone index (older uv, lagging index, or only a newer Python present), the install fails with "No interpreter found for Python 3.13.13 in managed installations or search path".

Loosening the pin to the `3.13` minor lets uv resolve to whatever stable 3.13.x is installed or downloadable. This stays safe from the rc-cache trap that motivated the original exact pin in #389: that trap was specific to the in-development 3.14 minor, where a stale cached 3.14.0rc1 outranked any final build. 3.13 is GA, so no prerelease sits above the latest stable patch, and uv does not pick prereleases by default. The Renovate rule that disables `python` minor and major updates still gates any jump to 3.14 behind review, and the Dockerfile keeps its exact `python:3.13.13-slim-bookworm` base image since that is a maintainer-controlled image unaffected by resolver lag.

Verified locally: `uv python find 3.13.99` reproduces the same "No interpreter found" failure class an end user hits, while `uv python find 3.13` and `uv run python --version` resolve cleanly against the loosened pin.


stickerdaniel:
* **#492** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/492?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/492?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #494, opened 2026-06-08
Fix/network filter coercion

johndimaster-ai:


stickerdaniel:
Thanks! Closing: stringified `network` arrays are already accepted since #747.

PR #505, opened 2026-06-11
fix(person): detect and accept incoming requests

### LINKED ISSUE #504: [BUG] Incoming connection requests misclassified as pending; accept path is English-only
### Setup

uv-managed local checkout on macOS (main @ df4a6e2), stdio/streamable-http, German LinkedIn session.

### What Happened

`connect_with_person` against a profile with an open incoming connection request (Accept/Ignore as primary top-card actions) returns `status: "pending"` with "A connection request is already pending for this profile." The expected state is `incoming_request`, which would accept the request and return `accepted`.

Two distinct defects:

1. **Misclassification as `pending`.** Incoming-request profiles have no Message button in the top card, so `findActionRoot` (extractor.py) walks up from sidebar compose anchors (mutual-connection cards carry `a[href*="/messaging/compose/"]` with aria-label) into a large container. That garbage root contains labeled anchors, so `has_labeled_action_anchor=True` and `detect_connection_state` returns `pending` before the incoming-request check is ever reached.
2. **Accept path is English-only.** Detection relies on `INCOMING_REQUEST_LABELS = {"en": ("Accept", "Ignore")}` and the click on `click_button_by_text("Accept")`. Non-English sessions (here: "Annehmen"/"Ignorieren") never match. The code carries a `TODO(locale)` noting no live incoming profile was available to probe; two are now available.

### Steps to Reproduce

1. Use a German-locale session with at least one open incoming connection request.
2. Call `connect_with_person(linkedin_username="<requester>")`.
3. Observe `status: "pending"` instead of `incoming_request`/`accepted`.

### Logs

```
{"status": "pending", "message": "A connection request is already pending for this profile.", "note_sent": false}
```

Live DOM probe of two incoming-request profiles (German locale) shows the top-card action row is three buttons sharing one parent: `button[aria-label]` (Accept) → `button[aria-label]` (Ignore) → `button[aria-expanded]` without aria-label (More). No data-* attributes, no accept deeplink URL, no embedded relationship JSON. A structural fingerprint (attribute presence + counts + order, per the AGENTS.md scraping rules) is therefore available to replace the text table as primary signal, with the locale table retained as fallback.

stickerdaniel:
Closes #504

`connect_with_person` could not handle profiles with an open incoming connection request. Incoming profiles render no Message button in the top card, so the compose-anchor action-root walk anchored on sidebar mutual-connection cards instead; their labeled anchors satisfied the pending signal and the incoming-request check was never reached. Detection and the Accept click were also bound to English text labels, so non-English sessions (verified live on a German session) always returned `pending`.

Detection is now fully structural and language-agnostic. A new signal fingerprints the Accept/Ignore action row, scoped to the top card (the first section of `main`): exactly three buttons in the smallest multi-button container around an unlabeled `button[aria-expanded]`, two of them carrying `aria-label` and preceding the expander, with no compose anchor, no invite anchor and no labeled anchor in the row. The match must be unique within that scope. All checks are attribute presence, counts and DOM order per the AGENTS.md scraping rules; no label text is read for any state. This signal is checked before the pending signal, so the sidebar mis-anchor no longer wins.

The Accept click targets the first labeled button in that row and only fires when the fingerprint matched, since clicking the second button would irreversibly ignore the request. There is deliberately no text-based fallback for either detection or the click: guessing a control by exact label text risks the wrong locale or an unrelated button on an irreversible action. The accepted state is verified with one settle retry because LinkedIn propagates it asynchronously; without the retry a successful accept could be reported as `send_failed`.

A new browser-DOM test module runs the real extractor JS against synthetic German fixtures in headless chromium, covering the live-replica incoming row plus hostile cases (video-player bars, expander-first ordering, extra buttons, a matching widget in a later section, follow-only, pending, connected). It auto-skips when no browser is installed, so CI is unaffected.

`uv run pytest` passes (544 tests, including 11 browser-DOM tests locally). `uv run ruff check .`, `uv run ruff format --check .` and `uv run ty check` are clean. Verified live against real LinkedIn: a read-only classification matrix over 10 profiles covering all seven connection states classifies every profile correctly with the new signal False on all non-incoming profiles, and an end-to-end `connect_with_person` against a real German-locale incoming request clicked Accept structurally and the profile transitioned to 1st-degree.

## Synthetic prompt

> connect_with_person misclassifies incoming-request profiles as pending (sidebar compose anchors poison the action-root signals) and its Accept detection/click is English-only. Replace it with a fully structural, locale-independent fingerprint of the Accept/Ignore action row (attribute presence, counts, DOM order, scoped to the top card and unique), checked before the pending signal, click the first labeled button only when it matches with no text fallback, verify the accepted state with a settle retry, and cover the fingerprint with real-DOM chromium tests plus unit tests.

### DIFF
--- a/linkedin_mcp_server/scraping/connection.py
+++ b/linkedin_mcp_server/scraping/connection.py
@@ -13,11 +13,12 @@ Per ``AGENTS.md`` Scraping Rules, classification logic relies on URL
 patterns and attribute *presence* — never on the values of locale-dependent
 text labels like "Connect", "Follow", or "1st".
 
-The single text-based fallback that remains is incoming-request detection
-(Accept/Ignore present in the top-card text). LinkedIn does not expose a
-distinctive URL or attribute for this state. Per the same rules, that
-fallback lives behind an explicit per-locale label table that is trivial
-to extend.
+Incoming-request detection is primarily structural: the Accept/Ignore
+action row is fingerprinted by attribute presence and element counts
+(see :attr:`ActionSignals.has_incoming_action_row`). LinkedIn exposes no
+URL for this state, so a text-based fallback (Accept/Ignore present in
+the top-card text) is retained behind an explicit per-locale label table
+for DOM variants the fingerprint does not cover.
 """
 
 from __future__ import annotations
@@ -38,12 +39,15 @@ ConnectionState = Literal[
 
 
 # Per AGENTS.md Scraping Rules: text-only signals must live behind an
-# explicit per-locale table. This table covers incoming-request detection
-# only — the one state without a structural URL/attribute signal we have
-# verified against the live DOM. Extend with additional ("xx", (a, b))
-# entries once the labels are confirmed against a real profile.
+# explicit per-locale table. This table is the incoming-request fallback
+# for DOM variants the structural action-row fingerprint does not match.
+# Extend with additional ("xx", (a, b)) entries once the labels are
+# confirmed against a real profile.
 INCOMING_REQUEST_LABELS: dict[str, tuple[str, str]] = {
     "en": ("Accept", "Ignore"),
+    # Verified live 2026-06-11 against two German-locale incoming-request
+    # profiles.
+    "de": ("Annehmen", "Ignorieren"),
 }
 
 
@@ -99,6 +103,19 @@ class ActionSignals:
     locale-dependent and not read; presence-on-an-``<a>`` is the
     locale-independent Pending signal."""
 
+    has_incoming_action_row: bool
+    """The top-card action row matches the incoming-request fingerprint:
+    exactly three ``<button>``s in the smallest multi-button container
+    around a ``button[aria-expanded]`` — two with ``aria-label``
+    (Accept, Ignore) preceding one unlabeled expander (More) — and the
+    container holds no compose anchor, no invite anchor, and no labeled
+    ``<a>``. All checks are attribute presence and structural counts per
+    the AGENTS.md Scraping Rules; no label values are read. Verified
+    live 2026-06-11 against two German-locale incoming-request profiles.
+    Computed independently of the compose-anchor action-root walk, which
+    finds no top-card root on incoming profiles (they have no Message
+    button) and would otherwise mis-anchor on sidebar cards."""
+
 
 def detect_connection_state(
     profile_text: str,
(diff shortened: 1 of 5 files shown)


stickerdaniel:
* **#505** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/505?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/505?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
@greptileai review

stickerdaniel:
Thanks, all three are real and fixed in f964d79:

- **P1 (fingerprint scans all of main):** the search is now scoped to the top card (`main.querySelector('section') || main.firstElementChild`), so a matching widget in a later section can't be detected or clicked. Added a DOM regression test (`test_matching_widget_outside_top_card_not_detected`). Verified live: across 10 profiles covering every state, `hasIncomingActionRow` is True only on the real incoming profile.
- **P2 (wrong-locale fallback click):** removed the locale-text fallback entirely. The structural row click is the only accept path; when the fingerprint doesn't match we report `send_failed` rather than clicking a button matched by exact text anywhere in the page.
- **P2 (fixture masks failures):** the `yield` is now outside the launch try/except, so an assertion failure is no longer converted into a skip.

@greptileai review

stickerdaniel:
Round 2 — both verified:

- **P1 (fingerprint scans all main):** the literal claim is addressed — the search is scoped to the top card (`main.querySelector('section') || main.firstElementChild`) at the top of `findIncomingActionRow`, and the regression test `test_matching_widget_outside_top_card_not_detected` proves a same-shaped widget in a later section is excluded. Since detection is now structural-only (below), I also hardened the click path: the fingerprint must match exactly one row in the scoped top card, so an ambiguous top card is a safe no-op rather than a guessed click.
- **P1 (fallback cannot accept):** real. Removed the locale text fallback from detection entirely. `detect_connection_state` is now fully structural, so detection and the accept action are consistent — a profile is `incoming_request` only when the fingerprinted row is present and clickable. This also drops the `profile_text` parameter and the locale label table.

Live matrix over 10 profiles still passes with the new signal False on every non-incoming profile.

@greptileai review

stickerdaniel:
Both comments on the latest commit are verbatim re-posts of the round-1/round-2 threads and are stale against the current code:

- **"Fingerprint scans all main":** `findIncomingActionRow` scopes to `const scope = main.querySelector('section') || main.firstElementChild` and iterates `scope.querySelectorAll('button[aria-expanded]')`, not `main`. It also requires a unique match in that scope. `test_matching_widget_outside_top_card_not_detected` covers a same-shaped widget in a later section.
- **"Fallback cannot accept":** the locale text fallback was removed. `grep` finds no `profile_text`, `_has_incoming_request_text`, or `INCOMING_REQUEST_LABELS` in `connection.py`/`extractor.py`; `detect_connection_state(signals)` is structural-only, so the path the comment describes no longer exists.

Not re-triggering, as that just re-anchors the same two threads. CI and the Greptile check are green.

PR #565, opened 2026-07-03
fix(bootstrap): show browser download progress

### LINKED ISSUE #533: [BUG] Installation Patchright browser download in background tool Error not showing progress
### Setup

-

### What Happened

-

### Steps to Reproduce

-

### Logs

```shell
-
```

ibtisam-12:
Instead of silently capturing subprocess output and discarding it, patchright install progress is now streamed line-by-line to logger.debug. This lets users see download progress (MB downloaded, percentage) when running with --log-level DEBUG.

CLI modes (--login, --status, --import-from-browser) additionally get live output via print() so the user is never left staring at a blank line.

The background task completion handler (_refresh_background_task_state) now logs the elapsed time, giving operators a clear signal that setup finished.

Fixes #533

### DIFF
--- a/btca.config.jsonc
+++ b/btca.config.jsonc
@@ -81,6 +81,13 @@
     },
     {
       "type": "git",
+      "name": "rich",
+      "url": "https://github.com/Textualize/rich",
+      "branch": "main",
+      "specialNotes": "Terminal rendering. Used for the browser-install progress bar; the bar glyph is hardcoded in progress_bar.py."
+    },
+    {
+      "type": "git",
       "name": "pythonCryptography",
       "url": "https://github.com/pyca/cryptography",
       "branch": "main",
--- a/linkedin_mcp_server/bootstrap.py
+++ b/linkedin_mcp_server/bootstrap.py
@@ -3,8 +3,11 @@
 from __future__ import annotations
 
 import asyncio
+import codecs
+from collections import deque
+from collections.abc import AsyncIterator, Callable, Iterator
+import contextlib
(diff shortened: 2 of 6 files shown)


stickerdaniel:
Thanks @ibtisam-12, merged. This is what I added on top before merging.

The streamed output now feeds a progress bar when stdout is a real terminal: percentage, transfer rate and ETA, read from patchright's own progress lines. What reaches the terminal is redacted and bounded, covering credentials in a proxy URL, control sequences, and the quoted response body of a failed download. A pipe, a dumb terminal, an unencodable glyph or a closed stdout falls back to plain lines rather than failing the install, and only log handlers writing to that same screen move above the bar.

Browser install progress

Verified with `uv run pytest` and real cold installs on a fresh profile. CI runs pytest with `-s`, where nothing repairs a closed stream between tests, and that surfaced a test of mine leaving `sys.stdout` unusable for everything after it. `tests/conftest.py` now fails any test that does the same.

## Synthetic prompt

> The patchright browser install downloads ~170 MiB on first run with no feedback (#533). Stream its output live, and when stdout is a real terminal render a progress bar (dots spinner, thin bar, percentage, transfer rate, ETA) driven by patchright's own progress lines. Fall back to plain lines for pipes, dumb terminals, unencodable glyphs and closed streams. Redact credentials and terminal control sequences, bound what is retained, drop quoted response bodies from failed downloads, and move only those log handlers that share the bar's screen above it.

Generated with Claude Opus 5 (high) and GPT-5.6 Sol (xhigh)

PR #566, opened 2026-07-06
fix(references): keep non-Latin reference labels

lfazliev:
## What

`clean_label` rejected any reference label without an ASCII alphanumeric
character (`[A-Za-z0-9]`), so Cyrillic and other non-Latin profile names —
common in RU/BY and other locales' people-search results — were silently
dropped. Relax the guard to the Unicode word-character check `[^\W_]`: a label
is kept when it has at least one letter or digit in any script, while pure
punctuation/symbol labels are still rejected.

## Tests

- `test_keeps_cyrillic_only_names` — a Cyrillic-only name survives cleaning.
- `test_rejects_punctuation_only_labels_across_scripts` — punctuation-only
  labels are still dropped.

## Synthetic prompt

> In `link_metadata.clean_label`, the `[A-Za-z0-9]` guard drops valid non-Latin
> (e.g. Cyrillic) profile names. Relax it to a Unicode word-character check so
> any-script letters/digits pass while punctuation-only labels stay rejected.
> Add regression tests.

### DIFF
--- a/linkedin_mcp_server/scraping/link_metadata.py
+++ b/linkedin_mcp_server/scraping/link_metadata.py
@@ -322,7 +322,13 @@ def choose_reference_text(
 
 
 def clean_label(value: str, kind: ReferenceKind) -> str | None:
-    """Normalize and compact a candidate label."""
+    """Normalize and compact a candidate label.
+
+    A label must contain at least one letter or digit in any script —
+    ``[^\\W_]`` matches a Unicode word character that is not an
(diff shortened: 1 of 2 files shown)


stickerdaniel:
Hey, thanks for this! Rebased onto main and pushed one commit on top: `[^\W_]` also lets the four Hangul fillers through, so an invisible-only label passed and shadowed the `aria_label` fallback. Excluded them.

PR #596, opened 2026-07-25
fix(auth): retire the browser profile on a new session

stickerdaniel:
Chromium mints `machine_id`, `session_id_generator_last_value` and `limited_entropy_randomization_source` into `Local State` once and then keeps them for the life of a profile directory. That directory was reused across logins and imports, so signing in with a second LinkedIn account handed LinkedIn the same device identity twice, alongside the previous account's Local Storage, IndexedDB and residual cookies. Verified on a real profile: it survived both `--login` and `--import-from-browser` untouched.

`rotate_source_profile()` moves the previous session aside before a new one is established, from the two functions that write `source-state.json` — `interactive_login` and `import_session_from_browser` — which together cover every CLI and bootstrap entry. Auto-import had no clearing at all before this. Retiring rather than deleting keeps a session recoverable if it turns out to have been fine, and `--logout` now clears those copies too, which is what the CLI already advertised.

Rotation is fail-closed. A half-moved tree would leave the next rotation splitting one session across two quarantines, so a failed move raises instead of logging and continuing. It also refuses while another process holds the profile: the auth root is a shared Docker mount, and rotating underneath a live Chromium corrupts both the old session and the new one.

Verified with 848 tests (nine new), ruff and ty. The rotation tests are mutation-checked: six of seven fail when the move is neutralised.

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,7 +308,7 @@ This opens a browser window where you log in manually (5 minute timeout for 2FA,
 - `--host HOST` - HTTP server host (default: 127.0.0.1)
 - `--port PORT` - HTTP server port (default: 8000)
 - `--path PATH` - HTTP server path (default: /mcp)
-- `--logout` - Clear all stored LinkedIn auth state, including source and derived runtime profiles
+- `--logout` - Clear all stored LinkedIn auth state, including source and derived runtime profiles and any retired sessions
 - `--timeout MS` - Browser timeout for page operations in milliseconds (default: 5000)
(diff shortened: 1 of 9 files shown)


stickerdaniel:
* **#596** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/596?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/596?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #598, opened 2026-07-25
fix(auth): coordinate the browser profile across processes

### LINKED ISSUE #597: [BUG] Concurrent server processes silently destroy the shared LinkedIn session
### Setup

- macOS 15 (Darwin 25.5.0), arm64, patchright 1.60.1 (Chromium 1228)
- Transport: stdio, several MCP client instances running at once
- Version: main as of #596

### What Happened

Every MCP client instance spawns its own server process, and each one launches its own Chromium against the same `~/.linkedin-mcp/profile`. Nothing coordinates them, and the result is silent session loss.

Measured with two processes writing a persistent cookie to one profile: in 3 of 3 runs the shorter-lived process's cookie survived and the longer-lived process's cookie was gone, even though the long-lived process closed last. That is the mechanism by which a working `li_at` disappears.

Two things make this invisible today:

1. `profile_in_use_by()` (`session_state.py:398`) reads Chromium's `SingletonLock`, but the default `chrome-headless-shell` never writes one. Verified: full Chrome for Testing in headless mode writes `SingletonLock`/`SingletonSocket`/`SingletonCookie`; the headless shell writes none. Since `headless` defaults to `True` (`config/schema.py:33`) and the shell is "the only binary the default headless scrape path launches" (`bootstrap.py:64`), the check only guards the headed `--login`.

2. The browser is not closed between successful tool calls — created on the first call and reused (`drivers/browser.py:406-436`), released only at lifespan teardown. So the profile is held for the whole process lifetime, and two long-lived clients overlap for hours.

`clear_auth_state()` (`session_state.py:606`) has no preflight at all and will delete the profile out from under a live browser.

### Steps to Reproduce

1. Create a valid session.
2. Start two server processes against the same `USER_DATA_DIR`.
3. Trigger a tool call in each.
4. Both open Chromium on the same profile without error; the last writer wins and the other session's cookies are lost.

### Logs

```shell
-
```

Related: supersedes #586 (closed as not-planned, same root cause); #330 reports the downstream symptoms on Windows.

stickerdaniel:
Closes #597

Every MCP client instance spawns its own server process, and each one launched its own Chromium against the same `~/.linkedin-mcp/profile`. Chromium does not refuse that: both browsers run, and the last one to close overwrites the other's cookies. Measured with two processes writing a persistent cookie, the longer-lived process lost its session in 3 of 3 runs. Driving one account from several browsers at once is also a detection signal in itself.

The existing `SingletonLock` check could not see this. Chromium writes that file only in full Chrome; the `chrome-headless-shell` the server launches by default writes none, so the check only ever guarded the headed `--login` path.

This adds a lease the project owns. One process holds the profile, keeps the browser open across tool calls as before, and hands over when another process announces interest through a second lock file. A waiter that arrives mid-scrape waits for that call to finish rather than for the owner to exit, so several clients take turns instead of one starving the rest. Nothing is killed and nothing is deleted.

Three properties are load-bearing, and each is pinned by a test. Contention raises `BrowserBusyError`, deliberately not an `AuthenticationError`: that class is routed into `invalidate_auth_and_trigger_relogin`, so misclassifying a lost race would retire the shared profile and destroy every other process's session. The lease is released only on a *confirmed* browser close, because `BrowserManager.close()` bounds its cleanup steps and swallows their failures, so it can return while Chromium is still alive. Rotation and logout refuse while any process is using the profile.

`fcntl.flock` is used directly rather than the `filelock` package: the pinned 3.29.1 unlinks the lock path before releasing it, which splits contenders across inodes — measured at 1122 mutual-exclusion violations across 900 critical sections. Windows uses `LockFileEx`, which has the shared/exclusive semantics the handoff signal needs and that `msvcrt.locking` lacks.

Coordination covers processes in the same runtime. `flock` does not carry across a Docker bind mount (verified against OrbStack), so the host/container boundary is documented rather than claimed.

Verified with 894 tests including a `subprocess`-based harness — every prior concurrency test ran in one event loop and could not have caught this — plus mutation tests confirming the two critical guards fail loudly when removed, and a two-process end-to-end run against a shared profile.

## Synthetic prompt

> Multiple MCP client instances each spawn their own server process and all open the same Chromium profile, silently corrupting the shared LinkedIn session. Add cross-process coordination so only one process uses the profile at a time, with a cooperative handoff so concurrent clients take turns rather than one blocking the rest. Make sure contention is never misreported as an authentication failure, and that rotation and logout refuse while the profile is in use.

Generated with Claude Opus 5


stickerdaniel:
* **#598** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/598?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/598?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
Both findings were real and are fixed in 9a601f7.

The exclusivity check released the lease before rotation and logout started moving files. It is now a context manager that holds it for the whole mutation, with a test that probes from a separate lease object mid-move.

The handoff signal was only probed after a tool call, so an owner that went idle never probed again. A background poll now drives both the handoff and the idle timeout. While adding the test for it I found the minimum hold window also applied when no tool call had run at all, which delayed handover from a completely idle owner; that window only exists to stop busy clients trading the browser back and forth, so it no longer applies there.

PR #602, opened 2026-07-26
fix(http): reject cross-origin requests to the MCP endpoint

stickerdaniel:
The `streamable-http` transport served requests from any origin. A website the user merely visits can resolve its own domain to this server's address, so the request arrives from the user's own browser, where a firewall cannot see it and neither can a bind address. It would have been answered with the logged-in LinkedIn session behind it. Measured against this server before the fix: `Origin: https://attacker.example` returned 200.

The cause is a default. FastMCP builds its session manager without transport security settings, and the SDK disables host and origin validation when none are supplied. Raising the FastMCP version alone changes nothing; the protection stays off until it is asked for.

Both halves of the guard matter, and the Host check carries most of the weight. A rebinding attack sends its own domain as *both* Host and Origin, so those agree and origin validation has nothing to object to. What gives the request away is that the Host names a server this one was never known by. Requests with no `Origin` at all stay served, which is every non-browser client.

Getting there took two corrections worth recording, because both passed the tests at the time. An earlier revision wildcarded the Host check off for exposed binds to keep them reachable, which accepted the attacker's Host and defeated the guard. The revision after that used FastMCP's `auto` policy, which only validates when the accepted connection landed on a loopback address, so a server bound to `0.0.0.0` and reached over its LAN address checked nothing at all. Every measurement that called itself an exposed bind had connected over `127.0.0.1`: the bind address was exposed, the connection was not. Strict validation applies either way.

Strict accepts localhost and the address a connection arrives on, which covers the documented flows including Docker, where a published port is reached at localhost. It does not accept a DNS name, so a machine name on the network or a public name in front of a reverse proxy is refused with 421. That is deliberate and now documented next to the exposed-bind example, along with the two ways out, rewriting the upstream Host or listing the name, because an unexplained refusal is how a guard like this ends up switched off.

Answering the interactive transport prompt used to leave the stored transport at `stdio` while an HTTP server ran. Both the bind warning and the gate that decides whether reading the local browser cookie is safe read that value, so a server listening on every interface looked private to them. The answer is now recorded and the configuration re-validated.

Only servers started with `--transport streamable-http` have a listener at all. The endpoint still has no authentication, so the bind address remains the boundary, and the warning about that fired only for a literal `0.0.0.0` or `::`, meaning a LAN address named outright warned about nothing. Loopback classification now resolves address literals through `ipaddress`, covering all of `127/8`, IPv4-mapped forms and case or trailing-dot spellings of `localhost`; names still fail closed, since DNS can point anywhere.

The wiring tests passed through both broken revisions, because they asserted on the keyword arguments handed to `mcp.run` and never asked what the server replies. `tests/test_transport_security.py` drives the real ASGI app instead, and runs every case against both a loopback and a non-loopback server address, which is the distinction the second bug hid behind. Reverting to `auto` fails nine of them, all on the exposed side.

Verified with 963 tests, plus end-to-end checks against real sockets reached over both loopback and a LAN address.

## Synthetic prompt

> The streamable-http MCP endpoint accepts requests from any origin, so a visited website can reach it by DNS rebinding and drive tools with the logged-in LinkedIn session. Enable host and origin validation on every connection, not just local ones, without breaking the documented Docker flow. Fix the exposed-bind warning so it covers every address reachable from outside the machine, and test it by asking a running server what it answers rather than by checking how it was configured.

### DIFF
--- a/.github/workflows/check-attribution.yml
+++ b/.github/workflows/check-attribution.yml
@@ -1,8 +1,10 @@
 name: Check Attribution
 
 on:
+  # Every pull request, not only those targeting main: a stacked pull request
+  # is based on the branch below it, and a trailer that slips through there
+  # reaches main when the stack merges.
   pull_request:
-    branches: [main]
 
 # Coding agents sign their work either with a Co-Authored-By trailer or by
 # claiming the author/committer field. Squash-merging collects the trailers of
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,7 @@ on:
     branches: [main]
   # Every pull request, not only those targeting main. A stacked pull request
   # is based on the branch below it, so filtering by base branch skipped CI
-  # entirely for the stacks this repository's workflow produces — the checks
-  # were not failing, they were absent.
+  # entirely for the stacks this repository's workflow produces.
   pull_request:
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,8 +4,11 @@ name: CI
 on:
   push:
     branches: [main]
+  # Every pull request, not only those targeting main. A stacked pull request
+  # is based on the branch below it, so filtering by base branch skipped CI
+  # entirely for the stacks this repository's workflow produces — the checks
+  # were not failing, they were absent.
   pull_request:
-    branches: [main]
 
 jobs:
   lint-and-check:
--- a/README.md
+++ b/README.md
@@ -347,6 +347,14 @@ docker run -it --rm \
 
 Runtime server logs are emitted by FastMCP/Uvicorn.
 
+The HTTP server answers requests addressed to `localhost` or to the address it
+is bound to, and refuses others with `421`. That is what stops a website you
+merely visit from pointing a domain at this server and using your LinkedIn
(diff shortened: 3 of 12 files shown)


stickerdaniel:
* **#603** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/603?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#602** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/602?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/602?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

PR #604, opened 2026-07-26
test(lease): run the profile lease on every platform

stickerdaniel:
The profile lease coordinates server processes through kernel file locks, and its Windows backend is a separate implementation built on `LockFileEx` rather than `flock`. CI runs on Ubuntu only, so that code had never executed anywhere. This adds a small matrix job running the two lease test files on macOS and Windows, which is the first time the Windows path runs at all. Writing tests to make that job meaningful surfaced defects in the backend and in the lease itself.

A forked child kept its parent's lease alive. `fork` duplicates the descriptor and the kernel lock survives as long as any copy is open, so a child that outlived its parent held the profile with nothing in its own state to explain why, and a third process stayed locked out until the child happened to exit. Resetting inside the accessors was not enough: a child calling only `handoff_requested()` still leaked, and so did one that touched the lease at all, which no amount of checking on entry could have caught. The reset now runs from a `fork` handler, so an inherited lease is discarded the moment the child exists, including after a nested fork. Closing is safe there but unlocking would not be, since both descriptions refer to the same open file and would release a lock the living parent still believes it holds.

On the Windows side, every `LockFileEx` failure was treated as contention, so a bad handle would have reported a permanently busy profile; only `ERROR_LOCK_VIOLATION` means that. That decision now has a name and is pinned on every platform, because a healthy run only ever produces error 33 and the matrix job would stay green if it regressed. Unlock had none of the `OSError` tolerance the POSIX path has, kernel32 was bound without prototypes, and binding it at import time could have taken the server down with an obscure error instead of the explicit refusal the lease already raises. A backend that raises also leaked the descriptor, which matters because the handoff probe swallows that error by design and the poller runs every second.

The lifespan skipped closing the browser when the handoff poller had died, leaving Chromium on the shared profile past shutdown, which is the corruption the lease exists to prevent reached through its own teardown.

One finding is about the tests rather than the code. The announce worker printed success even when the shared lock was never taken, because announcing degrades to a silent no-op by design. A platform whose shared locks are really exclusive would have passed every handoff test with only the first waiter holding anything, so the new matrix job would have gone green proving nothing. The worker now distinguishes the two cases and retries the genuinely transient race against the owner's probe, and the harness fails on worker EOF rather than spinning until the CI job's own timeout.

Config clamping and the poller's lifespan wiring had no tests at all; both are covered now.

Verified with 1044 tests and no skips, `profile_lease.py` at full line coverage, and mutation tests confirming each fix fails loudly when reverted. The Windows behaviour itself can only be proven by the new CI job, which passes.

## Synthetic prompt

> The profile lease's Windows LockFileEx backend has never run in CI, and config clamping plus the lifespan poller wiring have no tests. Add a platform matrix job running the lease tests on macOS and Windows, write the missing tests, and fix whatever the exercise turns up, verifying each fix with a mutation test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,3 +47,30 @@ jobs:
         run: uv run pytest --cov --cov-report=term-missing -n auto -v -s
 
       - run: uv cache prune --ci
+
+  # The lease coordinates server processes through kernel file locks, and the
+  # Windows backend is a different implementation from the POSIX one
+  # (LockFileEx rather than flock). The Ubuntu job above never executes it, so
+  # these two files run on every platform we ship on. Serial by choice: the
+  # tests spawn their own competing processes, and xdist would only add
+  # scheduling noise on top of that.
+  profile-lease-platforms:
+    name: Profile lease (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, windows-latest]
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+
+      - uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
+        with:
+          enable-cache: true
+
+      - run: uv sync --group dev
+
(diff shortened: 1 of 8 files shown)


stickerdaniel:
@greptileai review

PR #605, opened 2026-07-26
feat(config): Route browser traffic through a proxy

stickerdaniel:
There is no way to route the browser through a proxy. Users running the server on a VPS reach LinkedIn from an address that is obviously a data centre, and users whose compute sits in a different country than their account history sign in from the wrong place. The plumbing was already there: every launch ends at `launch_persistent_context`, which takes a `proxy` argument, and nothing was filling it.

`PROXY_SERVER` and `--proxy-server` take `scheme://host:port`, with the credentials in `PROXY_USERNAME` and `PROXY_PASSWORD`. There is no password flag: command-line arguments are readable by every other user on the machine. `PROXY_SERVER` does accept the combined `user:pass@host:port` form providers hand out, because the environment is not world-readable the way argv is; validation splits it apart. The same settings reach the login browser and the scraping browser, since a session created from one address and used from another is what trips a LinkedIn checkpoint.

Validation catches at configuration time what would otherwise surface as an unexplained failure much later. Chromium cannot authenticate to a SOCKS proxy at all — Patchright throws for it — so that combination is refused with an explanation. Credentials percent-encoded into the host (`user%3Apass%40host`) are refused too: `urlsplit` reads that as a plain hostname, so it would neither be split out nor kept from the logs, and Patchright resolves it to a nonsense host.

The larger half of the diff is making a proxy fault report itself as one. A dead proxy does not fail at launch; Chromium starts and the first navigation fails, which the auth checks read as a dead session. That raised an `AuthenticationError`, whose recovery moves the stored profile aside and starts a login through the same broken proxy — so a misconfigured proxy destroyed a working session. Wrong credentials are the sharp case: they produce no proxy error code at all, because Chromium retries the challenge until the navigation times out. Measured against a local authenticating relay. So with a proxy configured, a navigation that fails outright is now reported as a proxy fault, and only a recognised auth barrier — which requires a page to have loaded — may still conclude the session is invalid. Auto-import is skipped while a proxy is set, since a session taken from a local browser was created on the real address.

Credentials must not survive in a log. The password and the username are both kept out of the config repr that runs at DEBUG level, and out of the traces and logs on every failure path. The username counts because residential providers encode the account and zone in it. That took four boundaries: the feed check, the extractor's traces, its navigation-failure log, and the exception itself — everything downstream logs that, ending with FastMCP's own `logger.exception`, which writes the traceback before masking the client reply.

The docs say plainly that most people should not use this. LinkedIn advises against proxies and scores the addresses a session signs in from, so for a server running at home a proxy raises the risk it appears to lower. Where it genuinely helps, a WireGuard or Tailscale exit node on the user's own network beats a paid provider.

Verified with 1133 tests, and end to end against a local authenticating relay: Chromium loads through the proxy including the 407 handshake, an unreachable proxy reports the proxy rather than the session, and no credential reaches a log or `trace.jsonl`. Every guard is mutation-tested — reverting it makes a specific test fail.

## Synthetic prompt

> Add proxy support for the browser this server drives, configured through `PROXY_SERVER`/`--proxy-server` plus credentials, applied identically to the login and scraping paths. Validate at configuration time what Chromium cannot do, keep the credentials out of every log and trace, and make sure a proxy failure is never mistaken for an expired session — the recovery for that discards a working profile. Document honestly when a proxy is a bad idea.


stickerdaniel:
* **#605** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/605?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/605?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
@greptileai review

Issue #606, opened 2026-07-26
[FEATURE] Serve every MCP client from one shared browser


stickerdaniel:
### Feature Description

Every MCP client instance spawns its own server process. #597 established that they cannot share a Chromium profile safely, and #598 fixed that with a lease: one process owns the browser and hands over when another announces interest.

That makes concurrent use correct, but not cheap. Under sustained use each call arrives from a different client, the minimum-hold window has expired by the time the next one comes in, and ownership moves on every call. Each move closes and reopens Chromium, and every reopen runs `_feed_auth_succeeds` (`drivers/browser.py:153`), which navigates to `https://www.linkedin.com/feed/`.

So the steady state under real concurrent use is one extra LinkedIn request per tool call.

Instead, one long-lived process should own the browser and serve the others over an authenticated loopback endpoint. The stdio processes become thin proxies. All tool bodies then run in one process, which also makes the existing in-process lock effective everywhere.

### Use Case

Running several MCP clients at once is the normal case for this project, not an edge case. Request volume against LinkedIn is the thing worth minimising, so an extra feed navigation per call is the cost that matters here, ahead of the second or so of latency each reopen also adds.

Client configuration must not change. `uvx mcp-server-linkedin@latest` over stdio is the documented install (README.md:70-81) and the indirection has to be invisible.

### Suggested Approach

Three roles: a stdio frontend that proxies, a detached owner that holds the browser, and a supervisor that reaps the owner's process tree. Ownership is decided by a lock at the auth root, since sibling profile directories already share one lease.

Groundwork, each independently useful:

- #607 — attach middleware by what a server process does
- #608 — owner-only storage for the bearer token
- #609 — the ownership lock and the descriptor a client checks before attaching

Several findings constrain the design and are worth recording here:

- Cancellation is not forwarded by the pinned SDK, and killing the HTTP client does not stop the running tool (measured 0/4 across JSON and SSE, stateful and stateless). A liveness protocol has to be built rather than inherited.
- A child started with `start_new_session=True` is still killed when the MCP client shuts down; only a double-forked child survives.
- Chromium outlives its parent by 20 seconds or more after a kill, while the kernel frees the lock immediately. A free lock is therefore never proof the profile is idle.

This should not become the default until a killed client provably cannot leave the owner scraping while holding the browser. Until then it stays opt-in.

stickerdaniel:
[...]

PR #610, opened 2026-07-26
refactor(server): Attach middleware by what a process does

### LINKED ISSUE #607: [FEATURE] Attach middleware by what a server process actually does
### Feature Description

`create_mcp_server()` (`server.py:89`) attaches `SequentialToolExecutionMiddleware` and `UpdateNoticeMiddleware` unconditionally. Both of them only make sense for a process with a particular job, and today every server is assumed to have every job.

Serialization exists to protect the shared Chromium profile, so it belongs to a process that launches a browser. The update notice is appended to one tool result per process, so it belongs to a process an end user reads results from.

Those are not the same question, and a server that forwards its calls somewhere else answers no to both.

### Use Case

Groundwork for #606, where a process either drives the browser or forwards to one that does.

Attaching serialization to a forwarding process would have it take the profile lease it never uses, then wait out its own timeout or make the process that actually needs the lease wait for one held by a caller with no browser. Attaching the update notice to a shared long-lived process would show it to whichever client called first and to nobody after that, however many clients attach over its lifetime.

Worth doing on its own: it makes an implicit assumption explicit and testable before anything depends on it.

### Suggested Approach

Add a role to `create_mcp_server()`, defaulting to the current behaviour so existing callers are untouched. Attach serialization to the roles that drive a browser and the update notice to the roles a client reads from.

stickerdaniel:
Closes #607

`create_mcp_server()` attached `SequentialToolExecutionMiddleware` and `UpdateNoticeMiddleware` to every server, but each answers a different question about the process it runs in. Serialization protects the shared Chromium profile, so it belongs to a process that launches a browser. The update notice is appended once per process, so it belongs to a process an end user reads results from.

This adds a role to `create_mcp_server()` and attaches each middleware where it makes sense. The default is the historical server, so every existing caller is unchanged.

The failure this prevents is not hypothetical. Serialization on a process that does not drive a browser would take the profile lease it never uses and then either wait out its own `browser_wait_seconds` or hold the lease while the process that actually needs it times out. The update notice on a server shared by many clients would reach whichever called first and nobody after that, however long that server lives.

There are two roles rather than three. A forwarding role was written and then removed: nothing forwards yet, so it would have registered the local browser-backed tools and declined to serialize them, which is a worse starting point than not having the role. It returns with the code that gives it something to forward to.

Groundwork for #606; useful on its own for making an implicit assumption explicit and testable.

Verified with 1137 tests, four of them new, covering each role's middleware and that both serve the same tools. The middleware assertion iterates the roles rather than naming them, so one added later cannot quietly skip it.

## Synthetic prompt

> `create_mcp_server()` attaches the sequential-execution and update-notice middleware unconditionally, but one protects the shared browser profile and the other is user-facing. Add a role parameter so each is attached only where it belongs, keeping the current behaviour as the default, and test what breaks if either is attached in the wrong place.

### DIFF
--- a/linkedin_mcp_server/server.py
+++ b/linkedin_mcp_server/server.py
@@ -6,6 +6,7 @@ person profiles, company data, job information, and session management capabilit
 """
 
 import asyncio
+import enum
 import logging
 from typing import Any, AsyncIterator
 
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#612** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/612?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#611** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/611?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#610** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/610?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/610?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
@greptileai review

stickerdaniel:
Fixed in ca7e65e. Removed the forwarding role until something forwards: it registered the local browser tools and then skipped the middleware that serializes them, which is worse than not having the role. The middleware assertion now covers every role rather than naming them, so one added later cannot quietly skip it.

@greptileai review

PR #611, opened 2026-07-26
feat(security): Store local secrets owner-only on every platform

### LINKED ISSUE #608: [FEATURE] Store local secrets where only the current account can read them
### Feature Description

The server is about to keep a bearer token on disk (#606), and a token any local account can read is the same as no token.

On POSIX the existing helpers answer this: `secure_mkdir` and `secure_write_text` (`common_utils.py`) apply `0700` and `0600`. On Windows they do not. `os.chmod` there only toggles the read-only attribute; access is decided by the file's access control list, which nothing in the project touches. `harden_linkedin_tree` is an explicit no-op on Windows (`common_utils.py:43-51`).

Two gaps exist on POSIX as well, both verified:

- `secure_mkdir` applies its mode only to directories it creates. A directory that already exists keeps whatever permissions it was made with, so a `0600` token can sit inside a `0755` directory and look protected.
- `harden_linkedin_tree` deliberately stops outside a `.linkedin-mcp` directory, so a custom `USER_DATA_DIR` is not covered at all.

### Use Case

Directly needed by #606. Independently useful as the place any future local secret goes, rather than each caller reasoning about permissions again.

Worth stating plainly: on Windows a normal user profile is already closed to other non-administrator accounts, so this is defence in depth rather than the only thing standing in the way. It earns its place when the profile has been redirected, when a parent directory was created with wider permissions, or when the data directory sits outside the profile. Neither platform's mechanism keeps out an administrator or root, and no file permission ever has.

### Suggested Approach

One helper that makes the same promise on both platforms and verifies it rather than assuming it, since a filesystem that cannot store permissions accepts the call and keeps the old ones. On Windows that means replacing the access list rather than merging into it, so a permissive parent does not pass its grants down, and marking it protected so they are not inherited back afterwards.

Every path fails closed: if owner-only storage cannot be established, the caller must not write the secret.

The Windows half cannot be exercised on Linux or macOS, so the cross-platform CI job needs to cover it.

stickerdaniel:
Closes #608

The server is about to keep a bearer token on disk (#606), and a token any local account can read is the same as no token.

On POSIX the existing helpers answer this. On Windows they do not: `os.chmod` there only toggles the read-only attribute, access is decided by the file's access control list, and `harden_linkedin_tree` is an explicit no-op (`common_utils.py:43-51`).

Two gaps existed on POSIX too, both measured. `secure_mkdir` applies its mode only to directories it creates, so a `0600` token could sit inside a pre-existing `0755` directory and look protected. And `harden_linkedin_tree` deliberately stops outside a `.linkedin-mcp` directory, so a custom `USER_DATA_DIR` was never covered.

A third turned up during review and is the reason this is not just a Windows change: on macOS an access list sits alongside the mode rather than inside it, and `chmod` leaves it untouched. An auth root carrying an inheritable `everyone` entry produced a token file reporting exactly `0600` that everyone could still read, while hardening returned without complaint. Checking the mode was not checking access.

So this adds one helper that makes the same promise on both platforms and verifies it rather than assuming it. On Windows the access list is replaced rather than merged, so a permissive parent does not pass its grants down, and marked protected so they are not inherited back. A directory's entry is inheritable, which means a file created inside is owner-only from the instant it exists rather than from whenever something hardens it. On macOS the list is cleared through libc and read back; going through `chmod` and `ls` failed open when neither could be found on `PATH`.

Every path fails closed. A filesystem that accepts the call and stores nothing, which is what a share or a FAT volume does, is caught by reading the result back.

Worth being plain about the value: a normal Windows profile is already closed to other non-administrator accounts, so that half is defence in depth. It earns its place when the profile has been redirected, when a parent was created with wider permissions, or when the data directory sits outside the profile. Neither platform's mechanism keeps out an administrator or root.

The Win32 calls follow Microsoft's documentation, including that `SetNamedSecurityInfoW` returns its error code directly rather than setting the thread's last error, so checking `GetLastError` there would read a stale value and report a failure as success.

Verified with 1156 tests, 25 of them new. The four Windows tests cannot run on Linux or macOS, so the cross-platform CI job now covers them; they assert the resulting access list rather than mocking the calls.

## Synthetic prompt

> The daemon work needs a bearer token on disk. `os.chmod` does not protect a file on Windows, macOS access lists survive it, and the existing helpers do not harden a directory that already exists or one outside a `.linkedin-mcp` tree. Add a helper that makes a path readable only by the current account on both platforms, using the Win32 security APIs through ctypes on Windows and libc on macOS, verifying the result rather than trusting the call, and failing closed with no dependency on an external tool being findable. Extend the cross-platform CI job to run the Windows tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,14 +48,15 @@ jobs:
 
       - run: uv cache prune --ci
 
-  # The lease coordinates server processes through kernel file locks, and the
-  # Windows backend is a different implementation from the POSIX one
-  # (LockFileEx rather than flock). The Ubuntu job above never executes it, so
-  # these two files run on every platform we ship on. Serial by choice: the
+  # Everything here coordinates or protects state through the operating system
+  # rather than through Python, so each platform runs a different
+  # implementation: the lease locks with flock or with LockFileEx, and
+  # owner-only storage is a permission bit or an access control list. The
+  # Ubuntu job above executes neither Windows path. Serial by choice: the lease
   # tests spawn their own competing processes, and xdist would only add
   # scheduling noise on top of that.
-  profile-lease-platforms:
-    name: Profile lease (${{ matrix.os }})
+  platform-behaviour:
+    name: Platform behaviour (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -70,7 +71,7 @@ jobs:
 
       - run: uv sync --group dev
 
-      - name: Run profile lease tests
-        run: uv run pytest tests/test_profile_lease.py tests/test_profile_lease_integration.py
+      - name: Run platform behaviour tests
+        run: uv run pytest tests/test_profile_lease.py tests/test_profile_lease_integration.py tests/test_private_state.py
 
       - run: uv cache prune --ci
--- a/linkedin_mcp_server/common_utils.py
+++ b/linkedin_mcp_server/common_utils.py
@@ -60,6 +60,25 @@ def harden_linkedin_tree(path: Path) -> None:
             return
 
 
+def is_still_at(fd: int, path: Path) -> bool:
+    """Whether *fd* is still the file that *path* names.
+
+    Compared by device and inode rather than by counting links. A count catches
+    an unlink and misses a rename: the inode keeps its one link while the name
+    comes to mean a different file, so the count still reads as one. A path that
+    has since vanished counts as changed.
+
+    Only meaningful once whatever the caller wanted to establish is in hand.
+    Asked earlier, the answer can stop being true immediately afterwards.
+    """
+    held = os.fstat(fd)
+    try:
+        current = path.stat()
+    except OSError:
+        return False
+    return (held.st_dev, held.st_ino) == (current.st_dev, current.st_ino)
+
+
 def secure_write_text(path: Path, content: str, mode: int = 0o600) -> None:
     """Atomically write *content* to *path* with owner-only permissions.
 
--- a/linkedin_mcp_server/private_state.py
+++ b/linkedin_mcp_server/private_state.py
@@ -22,10 +22,11 @@ keeps out root or an administrator, and no file permission ever has.
 
 from __future__ import annotations
 
+import ctypes
+import ctypes.util
 import logging
 import os
 import stat
-import subprocess
 from pathlib import Path
 
 from linkedin_mcp_server.common_utils import secure_mkdir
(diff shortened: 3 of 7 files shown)


stickerdaniel:
* **#612** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/612?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#611** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/611?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/611?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#610** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/610?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

stickerdaniel:
@greptileai review

stickerdaniel:
Fixed in 70518b5. Reproduced first: an auth root carrying an inheritable `everyone` entry produced a token file reporting exactly 0600 that everyone could still read, and hardening returned without complaint. Checking the mode was not checking access. The list is now cleared and read back, and remaining entries are a refusal.

@greptileai review

stickerdaniel:
Fixed in 4a7a411. The access-list handling ran `chmod` and read `ls`, and both treated an unavailable executable as nothing to report. Measured with no usable PATH: hardening returned success while `everyone` could still read the token. Now through libc, so there is no executable to be missing.

@greptileai review

stickerdaniel:
Fixed in 1dd13a1. Reproduced first: the flag saying the library had been resolved was set before the lookup finished, so a concurrent first caller read an empty cache, which reads as "no access lists on this platform" everywhere it is used. That thread hardened a directory which kept an inherited everyone entry while reporting success. Now resolved under a lock with the flag set last, and a load that raises is no longer cached as an answer.

@greptileai review

stickerdaniel:
Fixed in 5b55b14. Reproduced first: loading a library raises a `ctypes.dlopen` audit event, the hook runs synchronously inside the resolution lock, and a hook reaching back into hardening left the resolving thread waiting on a lock it already held. Never finished. Reentrant lock now, with a test that deadlocks without it.

@greptileai review

stickerdaniel:
Fixed in df49d2b. Cross-model review found the owner-only guarantee was hollow in exactly the case it exists for.

The DACL was replaced, marked protected and verified. What was never read or set is the owner, and Windows grants an object's owner READ_CONTROL and WRITE_DAC implicitly, with no ACE saying so ([owner of a new object](https://learn.microsoft.com/en-us/windows/win32/secauthz/owner-of-a-new-object)). A redirected or shared profile can hold a directory owned by another account that merely grants this one WRITE_DAC: hardening succeeds, verification reports owner-only, and that account can widen the inheritable DACL straight back.

The window that matters is between hardening the daemon directory and writing the token into it. The write goes through a temporary file that inherits from the parent, so the secret would be on disk under permissions that other account chose, before `harden_file` runs.

Ownership is now taken alongside the DACL, and the owner is verified before the entries are, since a correct-looking DACL means nothing while somebody else can rewrite it. Neither this nor anything else here keeps out an administrator.

@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

PR #612, opened 2026-07-26
feat(daemon): Decide which process owns the browser

### LINKED ISSUE #609: [FEATURE] Decide which process owns the shared browser
### Feature Description

For one process to hold the browser on behalf of the others (#606), two things have to exist first: something that decides which process that is, and something a client can read to find it and check it is the right one to attach to.

The profile lease (#598) cannot do the first job. It is reference counted so a tool call, the browser singleton and a destructive helper can each hold it at once, and it is released the moment the browser closes so `--login` can take it. Ownership held that way would lapse whenever the browser happened to be shut, and holding it for a whole process lifetime would lock out every login.

### Use Case

Groundwork for #606, and the point where several measured constraints become code rather than notes.

### Suggested Approach

A separate lock, using the same kernel primitives with different rules: one holder, no reference counting, released only on exit. Placed at the auth root, because `auth_root_dir()` returns the profile's parent, so sibling profile directories already share one lease. An election scoped more narrowly would let two owners start and then compete forever for a lease only one can hold.

Two rules keep the lock honest, both measured:

- Release by closing the descriptor, never by unlocking it. `flock` belongs to the open file description, which an inherited copy shares, so unlocking frees it for a supervisor that inherited it too. Measured: unlock-then-close left the lock free while a live process believed it held it; close alone kept it held.
- Never read a free lock as proof the profile is idle. Chromium survived a kill of its parent by over 20 seconds while the kernel freed the lock immediately, so both facts are true at once.

Alongside it, a descriptor the owner publishes and a client verifies rather than trusts, since anything that can write to the auth root can edit it:

- the endpoint must be loopback before a token is sent anywhere;
- the profile path is compared exactly, since sibling profiles share one auth root and a client would otherwise be served the wrong session;
- the token lives in its own file named for the instance, so a descriptor and a token can never belong to different generations;
- configuration that decides what the shared browser is must match, compared through a digest keyed to the daemon token. Unkeyed it would cover the proxy password alongside otherwise guessable settings, which is an offline check for password guesses. Mismatches name fields, never values.

stickerdaniel:
Closes #609

For one process to hold the browser on behalf of the others (#606), two things have to exist first: something deciding which process that is, and something a client can read to find it and check it is the right one to attach to.

The profile lease (#598) cannot do the first job. It is reference counted so several callers in one process can hold it at once, and released the moment the browser closes so `--login` can take it. Ownership held that way would lapse whenever the browser happened to be shut, and holding it for a process lifetime would lock out every login.

So this adds a separate lock with the same kernel primitives and different rules: one holder, no reference counting, released only on exit. It lives at the auth root because `auth_root_dir()` returns the profile's parent, so sibling profile directories already share one lease; an election scoped more narrowly would let two owners start and then compete forever for a lease only one can hold.

Three properties come from measurements rather than reasoning.

The lock is released by closing the descriptor, never by unlocking it. A duplicate shares the lock rather than holding one of its own, so unlocking through any of them releases it for all. Measured on both platforms: unlock-then-close left the lock free while a live process believed it held it.

A free lock is never read as proof the profile is idle. A persistent headless Chromium survived a kill of its Python parent by over 20 seconds, while the kernel frees the lock at the instant of death.

Handing a held lock to a launched process works only on POSIX. Measured on a Windows runner: a child that received the handle through the documented `STARTUPINFO` mechanism and kept it open did not hold the lock once the parent had closed its handles and exited, and a third process took the byte range in 20 of 20 runs. Within one process a duplicate does keep it alive, which is why a single-process test agrees and the cross-process case does not. Both handoff entry points refuse on Windows and say the supervisor has to take the lock itself.

Alongside it, a descriptor the owner publishes and a client verifies rather than trusts, since anything that can write to the auth root can edit it. The endpoint must be loopback before a token is sent anywhere, and IPv6 literals are bracketed so an accepted endpoint is also a reachable one. The profile path is compared exactly, because sibling profiles share one auth root and a client configured for one would otherwise be handed the other's logged-in session. The instance identifier must be a UUID before it becomes a filename, and the token is opened without following links and confirmed to be a regular file, so neither the name nor a link at that name can redirect the read. The protocol version is compared, not merely recorded. Configuration deciding what the shared browser is must match, through a digest keyed to the daemon token: unkeyed it would cover the proxy password alongside otherwise guessable settings, which is an offline check for password guesses. Mismatches name fields, never values.

Three lock helpers in `profile_lease.py` become public so both callers share one implementation rather than duplicating it. Rename only, verified by diff.

Verified with 1290 tests, 135 of them new. Mutual exclusion, crash release, fork inheritance and the POSIX handoff are exercised with real subprocesses, since none of it can be shown inside one interpreter. The state directory, the identity it is keyed by and the aliases that must resolve to one browser were each reproduced as a split owner before being closed.

## Synthetic prompt

> Add the lock that decides which process owns the shared browser, and the descriptor a client reads to find and validate that owner. The profile lease cannot serve as the election lock, so build a separate one on the same primitives with one holder and no reference counting. Release it by closing rather than unlocking, never treat a free lock as proof no browser is running, and check on a real Windows runner whether an inherited handle actually carries the lock. Make the descriptor carry the exact profile path, a keyed configuration fingerprint and a token in its own instance-named file, and refuse anything that cannot be verified, including planted links and endpoints that are accepted but unreachable. Test mutual exclusion with real processes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,6 @@ jobs:
       - run: uv sync --group dev
 
       - name: Run platform behaviour tests
-        run: uv run pytest tests/test_profile_lease.py tests/test_profile_lease_integration.py tests/test_private_state.py
+        run: uv run pytest tests/test_profile_lease.py tests/test_profile_lease_integration.py tests/test_private_state.py tests/test_daemon_lock.py
 
       - run: uv cache prune --ci
--- a/linkedin_mcp_server/daemon_descriptor.py
+++ b/linkedin_mcp_server/daemon_descriptor.py
@@ -209,31 +209,36 @@ def daemon_state_root() -> Path:
     return (_account_home() / _APPLICATION_STATE_DIR / _DAEMON_DIR).resolve()
 
 
-def _directory_identity(path: Path, *, label: str) -> tuple[int, int]:
-    """Return the stable filesystem identity of a directory, creating it once.
+def _auth_root_identity(auth_root: Path) -> bytes:
+    """Return the stable filesystem identity of *auth_root*, creating it once.
 
-    For the process that is establishing state. The first daemon on a fresh
-    install must not key a missing path and then switch to an inode after
-    creating it, which would leave its own lock at an address nothing else
-    computes. ``secure_mkdir`` creates only what is missing and deliberately
-    leaves permissions on existing parents alone.
+    Created rather than merely inspected, because this runs in the process that
+    is establishing state: the first daemon on a fresh install must not key a
+    missing path and then switch to an inode after creating it, which would
+    leave its own lock at an address nothing else computes. ``secure_mkdir``
+    creates only what is missing and deliberately leaves permissions on
+    existing parents alone.
+
+    Device plus inode identifies the physical directory rather than one
+    spelling of it, which keeps case and Unicode aliases together on an
+    insensitive volume without conflating distinct directories on a sensitive
+    one. The auth root can be keyed this way because nothing rotates it; the
+    profile inside it cannot, and :func:`profile_identity` says why.
     """
-    canonical = path.expanduser().resolve()
+    canonical = auth_root.expanduser().resolve()
     secure_mkdir(canonical, mode=0o700)
     info = canonical.stat()
 
     if not stat.S_ISDIR(info.st_mode):
-        raise DescriptorError(f"The {label} is not a directory: {canonical}")
-
-    # Device plus inode identifies the physical directory, not one spelling of
-    # it. That keeps case and Unicode aliases together on insensitive volumes
-    # without conflating distinct directories on a case-sensitive volume.
+        raise DescriptorError(
+            f"The authentication root is not a directory: {canonical}"
+        )
     if not info.st_ino:
         raise DescriptorError(
             f"The filesystem does not provide a stable identity for {canonical}, "
             "so browser ownership cannot be coordinated safely"
         )
-    return info.st_dev, info.st_ino
+    return f"inode\0{info.st_dev}\0{info.st_ino}".encode("ascii")
 
 
 def profile_identity(profile: Path) -> str:
@@ -278,11 +283,6 @@ def profile_identity(profile: Path) -> str:
     except OSError:
         return spelled
     return f"under\0{info.st_dev}\0{info.st_ino}\0{name}"
-
-
-def _auth_root_identity(auth_root: Path) -> bytes:
-    device, inode = _directory_identity(auth_root, label="authentication root")
-    return f"inode\0{device}\0{inode}".encode("ascii")
 
 
 def daemon_dir(auth_root: Path) -> Path:
--- a/linkedin_mcp_server/daemon_descriptor.py
+++ b/linkedin_mcp_server/daemon_descriptor.py
@@ -1,0 +1,416 @@
+"""What a running daemon publishes about itself, and what a client checks.
+
+A client that finds a descriptor is about to send a bearer token to whatever
+address it names and then drive a logged-in LinkedIn session through it. So
+nothing here is taken on trust: the endpoint is checked to be loopback before
+the token goes anywhere, the profile path is compared exactly, and the
+configuration is compared through a fingerprint.
+
+Two fields exist because of measurements rather than tidiness:
+
+``profile_path``
+    ``auth_root_dir`` returns the profile's *parent*, so ``.../profile`` and
+    ``.../profile2`` share one auth root and therefore one lock. Without an
+    exact comparison here, a client configured for the second would attach to
+    the first one's browser and never notice.
+
+``config_fingerprint``
+    Two clients that disagree about ``headless``, the user agent or the proxy
+    are not interchangeable, because the owner's browser can only be one of
+    them. The fingerprint is keyed rather than a plain digest: ``proxy_password``
+    is part of what must match, and a plain hash over otherwise-guessable
+    configuration would be an offline check for password guesses.
+
+The token itself is deliberately *not* here. It lives beside the descriptor in
+its own file, named after the instance, and the descriptor carries only its
+digest. Fixed names would let a client pair one generation's descriptor with the
+next generation's token and read that mismatch as corruption rather than as the
+restart it is.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import hmac
+import json
+import os
+import secrets
+from collections.abc import Mapping
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any
+
+from linkedin_mcp_server.common_utils import secure_write_text, utcnow_iso
+from linkedin_mcp_server.config.schema import AppConfig, is_loopback_host
+from linkedin_mcp_server.private_state import harden_directory, harden_file
+
+#: Bumped when the shape below changes incompatibly. A client that does not
+#: recognise the value refuses rather than guessing at fields it cannot read.
+SCHEMA_VERSION = 1
+
+#: Bumped when the daemon's own protocol changes: the control routes, the call
+#: metadata, or the ping contract. Compatibility keys on this rather than on the
+#: package version, because the documented install is ``@latest`` and this
+#: project releases often. Matching exact versions would mean a manual shutdown
+#: after almost every release, which is precisely the friction the daemon exists
+#: to remove.
+PROTOCOL_VERSION = 1
+
+_DESCRIPTOR_FILE = "daemon.json"
+_DAEMON_DIR = "daemon"
+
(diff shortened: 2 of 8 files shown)


stickerdaniel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/612?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#612** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/612?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/612?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#611** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/611?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#610** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/610?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

stickerdaniel:
@greptileai review

stickerdaniel:
Both fixed in d55eeb8, and both reproduced first.

Path traversal: confirmed that `token_path(root, '../../../../etc/hosts')` resolved outside the daemon directory. The instance id is now required to be a UUID, checked where the descriptor is parsed as well as where it becomes a path.

Fork retention: confirmed that after the owner released and exited, an untouched fork child kept the lock held and `daemon_is_running` still reported true. Cleanup is now registered to run in the child immediately after a fork, matching the profile lease. Re-measured as false afterwards.

@greptileai review

stickerdaniel:
Fixed in 2b36651, and the Windows question is now measured rather than inferred.

Ran a standalone probe on a Windows runner. A child that received the handle through `STARTUPINFO.lpAttributeList['handle_list']` and kept it open did **not** hold the lock once the parent closed its own: a third process took the byte range in 20 of 20 runs. Within one process a duplicate does keep it alive (20/20 held), which is why a single-process test agrees and the cross-process case does not. `UnlockFileEx` through a duplicate releases the original's lock, 20/20, so the close-never-unlock rule is right within a process and irrelevant to transfer.

So the handoff is POSIX-only. Both entry points now refuse on Windows and say the supervisor has to take the lock itself; the transfer tests are POSIX-only and there is a Windows test for the refusal.

Also fixed: `adopt()` left the descriptor inheritable, so a later child inheriting descriptors kept the lock alive after the owner exited. An empty proxy password no longer compares equal to no password, since the launch options treat them differently. And the subprocess helper pins its working directory.

@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
Fixed in da49dbe, all three reproduced first.

The token was opened by path and followed what it found. The UUID check confines the filename, not where a link at that name points: a planted link whose target matched the digest was read and would have gone to the endpoint as the credential. Now opened without following links, confirmed a regular file, and read non-blocking so a named pipe there cannot hang the client inside `open()`.

`protocol_version` was parsed and never compared, so a descriptor from another protocol was accepted. Now enforced.

The lock registry was weak throughout, so a lock acquired and then dropped stayed held with nothing left to release it. Held locks are kept reachable until release; unheld ones stay weak.

@greptileai review

stickerdaniel:
Fixed in 7f400f1. The token read was hardened earlier; the descriptor read was not, and discovery touches it on every cold start. A named pipe left at that name blocked inside `open()` with no timeout and no error, and bytes that are not valid text escaped as a decoding error from both paths, which says nothing about the one distinction this module exists to make: absent, or present and untrustworthy. Both files now go through one reader.

@greptileai review

stickerdaniel:
Fixed in d1aae03. `O_NOFOLLOW` does not exist on Windows, so the flag fell back to zero, the open resolved the link, and the regular-file check then saw an ordinary file at the other end. Both daemon files were exposed, and the tests covering it are POSIX-only, so nothing said so. Windows now checks the entry before opening. That is check-then-open rather than atomic; closing the remaining gap needs the Win32 reparse-point flag, which is noted in the code for when the daemon actually runs there.

@greptileai review

stickerdaniel:
Fixed in 72fd711. Cross-model review found the ownership state was in the wrong place, and each fix for that exposed the next assumption underneath it. All of them were reproduced as a real split owner first.

The lock lived under the auth root, which is the parent of `USER_DATA_DIR` and therefore whatever the user configured: acquiring turned a user-owned 0755 directory into 0700, and pointed at `/tmp` that is an attempt to chmod `/tmp`. Moving it under the account's own directory then needed the key to actually identify one directory. It did not: macOS keeps the caller's case through `resolve()` and `normcase` is a no-op there, so two spellings of one profile hashed differently and both processes took the lock. Device and inode fixed that, and cover Unicode aliases too.

Three more followed. `Path.home()` reads the environment, so a launcher setting `HOME` for one process gave it a private root and the same profile two owners. A missing auth root was keyed as a path and then created, so the first daemon's own lock sat at an address nothing else computed. And the state root itself had to be refused as a symlink, since the lock is taken by path and retargeting the link between two acquisitions splits the election again.

The descriptor now carries the profile's device and inode next to its path, hence the schema bump: comparing paths meant a client naming the same profile in another case was refused a browser the lock had already agreed was one. Comparison creates nothing, so assembling that refusal cannot leave a profile directory behind.

Separately: `auto_import_from_browser` is tri-state where `None` means enabled, so an owner without the flag and a client with `--auto-import` disagreed on a field where they behave identically.

One thing this deliberately does not solve, recorded in the module for whoever wires it up in #606: two containers sharing a bind-mounted auth root do not share this state, because each has its own home. Measured, both took the lock. Docker either stays on the per-operation lease or gets a shared state directory of its own.

@greptileai review

stickerdaniel:
Fixed in 1c17352. A second review pass found the symlink check from the previous round was aimed at the right problem and solved none of it.

A home reached through a link is an ordinary POSIX layout, so refusing one refuses a working machine. Worse, the check looked only at the two directories below the home and never at the home itself, which is the layout it was most likely to meet. Measured with a linked home retargeted between two acquisitions: both processes took a lock, on different inodes, at identical paths. Resolving is what the goal actually asked for, and a normal layout keeps working.

Also refused rather than created: a home that is not there. Creating it would shadow one that is merely not mounted yet, and that daemon would key its state under the empty directory while everything started after the mount keys it under the real one.

Two things from that review I did not change, both verified first. Replacing the auth root under a running owner does let a second one start, but the existing profile lease behaves identically, so it is a property of any lock addressed by path rather than something this introduces. It is now stated in the docstring instead of hidden behind a check that only appears to close it. And the report that `__init__` caches a path `try_acquire()` recomputes does not hold: both use the same `self._path`.

@greptileai review

stickerdaniel:
Fixed in 5e51b3a. Two more from the final review, both reproduced first.

Keying the descriptor's profile on device and inode was wrong about what a profile directory is. Every path that establishes a new session rotates the old one into quarantine and Chromium creates a fresh directory in its place, so the inode changes at the first login or import. Measured: a descriptor published before that login stopped matching its own profile afterwards, while its owner went on holding the lock and no client could recognise the daemon it was meant to attach to. The auth root above it is not rotated, which is why the lock can key on an inode and this cannot. The profile is now identified by the parent's inode plus its own name, with the name read from the parent's directory listing rather than taken from the caller, which is what makes `Profile` and `profile` agree where the filesystem says they are one directory.

Adoption took the caller's word for two things it never checked: that the descriptor belongs to this auth root's lock file, and that it carries the lock at all. Measured: a descriptor from a different auth root was adopted without complaint, so the process reported it owned a browser nobody had elected it for while a contender took the real lock alongside it. It is now compared by device and inode, and the lock is asked for rather than assumed. Asking is not acquiring twice, since a process that already holds it is granted it again against its own open file description.

The Windows ownership finding from the same review belongs to #611 and was fixed there.

@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
GitHub closed this when #611 merged and its base branch was deleted, and it cannot be reopened. Continued in #617 with the same branch and the same commits, retargeted at main.

PR #622, opened 2026-07-28
feat(daemon): Find a running owner worth attaching to

stickerdaniel:
Running several MCP clients at once means several server processes on one Chromium profile. #598 made that safe with a file lease, but ownership moves on nearly every call, and each move closes and reopens the browser, which revalidates the session against `/feed/`. The steady state is one extra LinkedIn request per tool call. Fixing it means one long-lived process owning the browser and serving the others over authenticated loopback HTTP, which is #606.

This is the first piece: reading the descriptor a daemon publishes and deciding whether this client may talk to it. It elects nobody, spawns nothing, takes no lock and forwards no calls, so with `daemon_enabled` off it changes nothing for anyone.

The result is a typed state rather than an optional attachment, because "no owner to attach to" covers four situations that call for different responses: nothing published, a descriptor for another runtime or profile or configuration, one that cannot be trusted, and one that is usable. Collapsing them is how a second Chromium ends up on a live profile. The single rule underneath is that a file says nothing about whether the process that wrote it still exists: only the lock settles who owns the browser, and only connecting settles whether anything is listening. A descriptor that passes every check is still just a file an owner may have left behind when it crashed.

`ServerRole` moves to its own module. Role has to decide behaviour down in `dependencies`, which `server` reaches only through the tool modules, so importing it back would close a cycle.

Verified with the full suite plus mutation checks on each refusal, confirming the test fails when the corresponding check is removed.

### DIFF
--- a/linkedin_mcp_server/config/loaders.py
+++ b/linkedin_mcp_server/config/loaders.py
@@ -117,6 +117,7 @@ class EnvironmentKeys:
     IMPORT_FROM_BROWSER = "IMPORT_FROM_BROWSER"
     AUTO_IMPORT_FROM_BROWSER = "AUTO_IMPORT_FROM_BROWSER"
     EAGER_FULL_CHROMIUM = "EAGER_FULL_CHROMIUM"
+    DAEMON_ENABLED = "DAEMON_ENABLED"
 
 
 def is_interactive_environment() -> bool:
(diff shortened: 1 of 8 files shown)


stickerdaniel:
* **#622** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/622?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/622?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
@greptileai review

PR #624, opened 2026-07-28
fix(runtime): Stop reading other containers as our own

### LINKED ISSUE #621: [BUG] Container detection false-positives on hosts running unrelated Docker containers (live repro of #529, never fixed)
## What happened

On a normal Linux development machine, running the server natively (not inside any container), `--login` completes successfully and writes a valid session (`cookies.json`, `source-state.json`, and a populated Chromium profile all under `~/.linkedin-mcp/`). Every tool call afterward still fails with:

```
No valid LinkedIn session is available in Docker. Run --login on the host machine to create a session, then retry this tool.
```

This happens even though the session is genuinely present and valid, and persists after clearing stale `SingletonLock` / `profile.lock` files and restarting the server process.

## This is the same bug as #529, which was never actually fixed

#529 diagnosed this exact root cause (same functions, same lines, same fix direction) as a code-audit finding, then was self-retracted by the reporter as "filed in error" before any patch landed. I independently hit the live, reproduced version of it today and confirmed the flagged code is unchanged in the current release (`mcp-server-linkedin` 4.20.0, `linkedin_mcp_server/session_state.py`, same lines as #529 described). Re-opening as a confirmed, real-world repro rather than only an audit finding.

## Root cause

`_is_container_runtime()` (https://github.com/stickerdaniel/linkedin-mcp-server/blob/main/linkedin_mcp_server/session_state.py#L158-L184) treats any occurrence of the strings `"docker"`, `"containerd"`, `"kubepods"`, `"podman"`, `"libpod"` anywhere in `/proc/1/mountinfo` or `/proc/self/mountinfo` as proof the current process itself is running inside a container:

```python
for probe in (
    Path("/proc/1/mountinfo"),
    Path("/proc/self/mountinfo"),
):
    if _path_contains_markers(probe, markers) or _root_mount_uses_overlay(probe):
        return True
```

`mountinfo` lists every mount visible in the process's mount namespace, not just its own root. On a normal (non-containerized) host that simply has the Docker daemon running with any containers active for unrelated purposes, those containers' overlay filesystems show up as ordinary mount entries under `/var/lib/docker/rootfs/overlayfs/<id>`, so the substring `docker`/`overlay` is present in the file regardless of whether the *current process* is containerized. This is an extremely common developer setup: any machine running Docker Compose for local Postgres/Redis/Kafka/etc. alongside this server will trip it.

The correctly-scoped check already exists right next to the broken one: `_root_mount_uses_overlay()` (lines 199-219) checks whether the process's own root (`/`) mount is of type `overlay`, the actually meaningful signal. The `_path_contains_markers(probe, markers)` call on the mountinfo probes is redundant with it and is what produces the false positive.

Confirmed on the affected machine:
- `/.dockerenv`, `/run/.containerenv`, `/run/containerenv` — none exist
- `/proc/self/cgroup` — a normal user-session cgroup path, no docker/containerd controller-path membership
- `findmnt /` — real `ext4` root filesystem, not overlay
- `/proc/self/mountinfo` — contains `docker`/`overlay` substrings only because of unrelated Docker containers running on the same host for other local services

So `get_runtime_id()` (line 132) appends `-container`, `get_runtime_policy()` (`bootstrap.py`) resolves to `RuntimePolicy.DOCKER`, and `ensure_tool_ready_or_raise()` routes into `_raise_if_docker_auth_missing()`, which never looks at the real, valid, already-authenticated profile at all since that check only runs in `RuntimePolicy.MANAGED`.

There's no environment variable or config flag to override the detected runtime policy, so once this misfires it's permanent: every tool call fails with a misleading "please log in" message even though login already succeeded, with no way to unblock short of patching the source.

## Local validation

Removing the `_path_contains_markers(probe, markers)` half of the mountinfo check (keeping `_root_mount_uses_overlay(probe)`) and restarting the server process resolved it immediately in our environment, consistent with the fix #529 already proposed.

## Suggested fix

Per #529's own suggestion, which still applies: drop the unanchored `docker`/`containerd` substring scan of mountinfo; restrict positive container evidence to the explicit marker files (`/.dockerenv`, `/run/.containerenv`, `/run/containerenv`), line-anchored cgroup controller-path membership (not a raw substring scan), and the existing root-overlay check. As a belt-and-suspenders improvement, an explicit override (e.g. `LINKEDIN_MCP_RUNTIME_POLICY=managed`) would let affected users unblock themselves immediately without waiting on a release.

Happy to open a PR with this change if useful.

— Cipher Solutions (https://ciphersolutions.dev)

stickerdaniel:
Closes #621

On a Linux machine that is not a container but simply runs a Docker daemon for other things — a local Postgres, a Redis — every tool call answered "No valid LinkedIn session is available in Docker. Run --login on the host machine" while a perfectly valid session sat on disk. No flag or variable reached that decision, so the only way out was editing installed source.

Detection searched `/proc/*/mountinfo` and `/proc/*/cgroup` for the substrings `docker`, `containerd` and friends anywhere in the file. Those files describe what the mount namespace can *see*, not what this process *is*: other containers' overlay mounts are ordinary entries on their host. The correctly scoped check already sat next to the broken one.

Every signal now describes this process. The root mount is judged from its own `/` line, using the fields the kernel controls rather than the mount source, which is a label the other end chooses. Cgroup membership is matched as whole path segments, so `docker.service` reads as what it is — the daemon, running on a host — and systemd's `app-docker\x2ddesktop.scope` no longer counts.

`LINKEDIN_MCP_CONTAINER=false` (or `true`) overrides the result. A heuristic over other people's kernels will be wrong somewhere, and the point of this bug is that being wrong was unrecoverable.

LXC and LXD are now detected, which they never were on any previous version.

## Verification

Reproduced and then re-verified on real Linux — an Ubuntu VM with Docker, containerd, LXC 5.0.3 and a local NFS export — not simulated `/proc`. That testing found four real problems in the first attempts, two of which would have shipped:

- A containerd container using the `native` snapshotter bind-mounts its rootfs onto the host filesystem, so checking for an overlay *type* alone read it as a host. False negatives are the worse direction: the container would then hunt for a browser keychain that does not exist.
- The root check initially read the mount *source*. On NFS that is a server-chosen export path, so a workstation whose root came from `nas:/var/lib/containers/workstations/alice` classified as a container — the original failure, reintroduced by a different route. Verified fixed against a real NFSv3 mount with that exact shape.
- A `docker-` cgroup prefix matched any host service, so an ordinary `docker-backup.scope` counted as a container. Instance segments now require a full container id, measured at 64 hex characters.
- `/run/systemd/container` looked like the authoritative signal and is not one for this question. An OrbStack Linux machine reports `lxc` there while being a full system with its own systemd, a persistent disk and a desktop-class user. Trusting it classified that machine as a container and blocked every tool call. The marker answers whether a boundary exists; what matters here is whether a browser and a keychain sit on the other side, and no single flag says that.

Confirmed on real Linux: plain Docker, the repo's own image, `--cgroupns host`, the vfs storage driver, rootless Podman with a real 64-hex libpod scope, raw containerd with the native snapshotter, and a genuine nested LXC container all detect as containers. A host running an unrelated Redis container, a host scope named `docker-backup.scope`, an NFS-rooted host, and the OrbStack machine hosting that LXC container all stay hosts.

Two limits worth stating rather than glossing: systemd-nspawn is covered by fixtures only, because the sandbox denies the cgroup mount nspawn needs, and a host whose `/` is genuinely an overlay still reads as a container — as it did before this change. `LINKEDIN_MCP_CONTAINER` is the lever for both.

`test_get_runtime_id_ignores_non_root_overlay_mounts` already asserted the invariant this bug violated, but used a path under `/var/lib/containers` with no marker word, so it passed either way. It now has a sibling using a real Docker host's mountinfo, and both fail if the old scan is restored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,10 @@ while a container is running.
 - *First tool call with no session*: if a locally logged-in browser has a live LinkedIn session, the server auto-imports it (see `AUTO_IMPORT_FROM_BROWSER` / `--auto-import`) instead of forcing a manual login. On macOS the keychain may prompt once for Safe Storage access. If no importable browser session exists, it falls back to opening a login window and waits up to `LOGIN_INLINE_WAIT` seconds (default 25, max 45; `--login-inline-wait`) so a quick sign-in resolves in one call. If the wait elapses, the tool returns a pending signal and the model retries in about 30 seconds. Neither the auto-import nor the inline wait applies under Docker or when the server is bound to a non-loopback HTTP host; create the session on the host with `--login`.
 - Users on slow connections may need higher values for either.
 
+**Told to run `--login` on the host when you already did:**
+
+- If tool calls answer "No valid LinkedIn session is available in Docker" on a machine that is *not* a container, the runtime was misdetected. This happened on Linux hosts running a Docker daemon for unrelated services. Set `LINKEDIN_MCP_CONTAINER=false` to override the detection; `true` forces the opposite.
+
 **Using a proxy:**
 
 > **Most people should not use one.** LinkedIn's own guidance for reducing
@@ -279,6 +283,10 @@ On startup, the MCP Bundle starts preparing the shared Patchright Chromium brows
 - *Entire tool calls timing out* (e.g. multi-section profiles, cold-start Chromium, slow containers): increase the per-tool execution timeout — `--tool-timeout 300` or `TOOL_TIMEOUT=300` (seconds, default 180).
 - *First tool call with no session*: if a locally logged-in browser has a live LinkedIn session, the server auto-imports it (see `AUTO_IMPORT_FROM_BROWSER` / `--auto-import`) instead of forcing a manual login. On macOS the keychain may prompt once for Safe Storage access. If no importable browser session exists, it falls back to opening a login window and waits up to `LOGIN_INLINE_WAIT` seconds (default 25, max 45; `--login-inline-wait`) so a quick sign-in resolves in one call. If the wait elapses, the tool returns a pending signal and the model retries in about 30 seconds. Neither the auto-import nor the inline wait applies under Docker or when the server is bound to a non-loopback HTTP host; create the session on the host with `--login`.
 - Users on slow connections may need higher values for either.
+
+**Told to run `--login` on the host when you already did:**
+
+- If tool calls answer "No valid LinkedIn session is available in Docker" on a machine that is *not* a container, the runtime was misdetected. This happened on Linux hosts running a Docker daemon for unrelated services. Set `LINKEDIN_MCP_CONTAINER=false` to override the detection; `true` forces the opposite.
 
 </details>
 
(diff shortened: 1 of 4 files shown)


stickerdaniel:
* **#624** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/624?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/624?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

PR #628, opened 2026-07-29
feat(daemon): Elect one owner and serve it over loopback

stickerdaniel:
The second piece of #606, after #622 shipped the read-only half. This one starts a process: a detached owner that holds the daemon lock for its lifetime, drives Chromium, and serves an authenticated endpoint on a loopback port. `daemon_enabled` is still off, and the frontend still runs its own tools, so nothing changes for anyone who leaves the flag alone.

Election is two sides rather than one function, because the process that wins the lock is not the process that serves: `cli_main` runs the stdio server blocking, so the winner cannot also serve owner HTTP. On POSIX the frontend takes the lock first and hands a duplicate to the child, which closes the window in which the position would look free. Windows has no fd transfer, so the frontend takes no lock and the child competes for it. The parent then releases its own descriptor before serving anything, and that line is load-bearing: both descriptors refer to one locked open file description, so a frontend that kept its copy would keep the daemon lock alive after the owner died and lock out every recovery.

The owner binds the port itself, proves the endpoint answers a real authenticated request, and only then publishes and signals ready. Publishing earlier advertises a server that refuses tokens.

Version skew is decided here rather than left implicit. `package_version` was published and never compared, so with `@latest` mandatory a freshly updated frontend could be served by a months-old owner indefinitely. A newer frontend now asks the owner to stand down and elects a replacement; an older one attaches, because restarting a shared browser to satisfy one stale client is the worse trade.

## What running it found

Twenty-two defects, each reproduced before it was fixed. Three were ways the daemon's credentials could leave the machine or reach the wrong process:

- **the working directory could hijack the owner.** `python -m` puts the inherited cwd first on `sys.path`, so a workspace containing `linkedin_mcp_server/daemon_owner.py` was imported in preference to the installed package. That code received the inherited lock descriptor and the entire configuration on stdin, proxy password included, merely because an MCP client happened to start there. `-P` alone does not close it — `PYTHONPATH=.` puts the workspace back — so the spawn runs in isolated mode.
- **loopback requests honoured `HTTP_PROXY`**, including for `127.0.0.1`. Reproduced against a capture proxy, which received the absolute URL and the bearer token for a logged-in LinkedIn session in full.
- **the child inherited `PROXY_PASSWORD` and `PROXY_SERVER`**, the latter of which accepts embedded credentials in its documented form.

Six were ways it could hang, wedge, or double up. Four of those defeated a bound that existed to prevent exactly that:

- **the handshake cleanup undid its own timeout.** `child.stdout.close()` waits on the reader thread's I/O lock, so it blocked for as long as the child stayed silent: 29.27 s against a 30 s sleeper, unbounded against a child that never speaks.
- **the configuration write ran before any budget applied**, and a child that never read it was then left holding the inherited lock forever. Bounding the write fixed the visible hang; the lock stayed wedged until the child was stopped too.
- **the owner's own timeouts ended the wait but not the process.** `asyncio.run` then cancels the pending task and waits, unbounded, for that cancellation — and `close_browser` suppresses cancellation by design, with an unbounded cookie export in front of it.
- **a handshake timeout was read as failure**, so the frontend would have driven its own browser against the profile the still-starting child was about to open.
- **Windows was never detached.** `start_new_session` is POSIX-only; CPython names the parameter `unused_start_new_session` there.

And one that was silently wrong rather than stuck: **the owner ignored the browser mode it was handed.** That lives in a module global defaulting to headless, which `_make_browser` reads instead of the configuration. An owner started with `--no-headless` published a fingerprint saying so, the frontend compared it and attached, and the browser opened headless anyway — no window, no error.

The rest: retrying stopped after one attempt whenever a wait budget was given, which also meant a frontend that lost the race never took over when the winner died; a crashed owner's descriptor was handed back as attachable; a version turnover then spent 90 s re-reading the descriptor it had just rejected; the stand-down route compared tokens as strings and answered 500 rather than 401 on any non-ASCII character.

Ten tests that passed against broken code were rewritten until they failed. `assert attempts >= 1` held for an implementation that gave up immediately — strengthening it exposed the retry bug above. A probe count held when nothing was probed at all. One test killed the very child whose survival was the defect.

## Measured

Eight clients electing concurrently produce exactly one owner. A child that dies after adopting the lock leaves it free, and so does one stopped before it could adopt. The owner survives SIGKILL to its client's whole process group and reparents to pid 1. Five rounds of kill-and-re-elect never produced two browsers. Twenty failed spawns leak no descriptor and no lock. A wedged shutdown exits with the lock free.

Not measured, and stated in the code: Windows detachment and the Windows handshake release, both of which have only the documented mechanism behind them rather than an observed outcome; and Linux under a supervisor that kills a cgroup rather than a process group. These are gates before the flag is turned on by default; none is a regression while it is opt-in.

### DIFF
--- a/linkedin_mcp_server/cli_main.py
+++ b/linkedin_mcp_server/cli_main.py
@@ -14,6 +14,7 @@ from linkedin_mcp_server.bootstrap import (
 from linkedin_mcp_server.core import AuthenticationError
 from linkedin_mcp_server.authentication import clear_auth_state
 from linkedin_mcp_server.config import get_config
+from linkedin_mcp_server.config.schema import AppConfig
 from linkedin_mcp_server.drivers.browser import (
     experimental_persist_derived_runtime,
     close_browser,
@@ -282,6 +283,48 @@ def profile_info_and_exit() -> None:
     sys.exit(1)
 
 
+def _elect_daemon_owner_if_enabled(config: AppConfig) -> None:
+    """Make sure one shared owner is running, when the daemon is switched on.
+
+    Off by default, and inert when off. What it establishes when on is the half
+    of the daemon that has to work before forwarding can be built: exactly one
+    owner process per profile, reachable, with the lock held by the process that
+    actually serves.
+
+    This process still runs its own tools. Until the forwarding role exists, an
+    enabled daemon therefore costs an idle owner beside the ordinary server and
+    buys nothing, which is why the flag stays experimental and off. Nothing here
+    can make things worse for a user who leaves it alone.
+
+    Never fatal. Every failure inside is a reason to serve this client the way
+    the server always has, and a client that could not start because a shared
+    browser could not be elected would be a worse outcome than not sharing one.
+    """
+    from linkedin_mcp_server.daemon import daemon_would_be_used
+
+    if not daemon_would_be_used(config):
+        return
+
+    try:
+        from linkedin_mcp_server.daemon_election import obtain_owner
+        from linkedin_mcp_server.session_state import auth_root_dir
+
+        profile = get_profile_dir()
+        outcome = obtain_owner(auth_root_dir(profile), profile, config)
+    except Exception:
+        logger.warning("The shared browser owner is unavailable", exc_info=True)
+        return
+
+    if outcome.worth_connecting:
+        logger.info("A shared browser owner is running")
+    else:
+        logger.warning(
+            "No shared browser owner could be started (%s); this server will "
+            "drive its own browser",
+            outcome.attachment_lookup.state.value,
+        )
+
+
 def get_version() -> str:
     """Get version from installed metadata with a source fallback."""
     try:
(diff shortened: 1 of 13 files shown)


stickerdaniel:
* **#628** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/628?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/628?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
@greptileai review

stickerdaniel:
### Merge activity

* **Jul 29, 2:04 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/628).
* **Jul 29, 2:04 PM UTC**: @stickerdaniel merged this pull request with [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/628).

PR #634, opened 2026-07-31
feat(daemon): Read the server role process-wide

stickerdaniel:
Assembling a server takes the role as an argument, and that settles
everything the assembly decides. The auth gates are not like that:
whether a login window may open is decided in bootstrap, reached from a
tool body with no server to ask. A detached owner has no terminal and no
desktop session, so it has to be able to find out what it is.

So the role is also process state, in server_role, which is already
import-free for exactly this class of reader.

Unclaimed is its own state rather than DIRECT. With DIRECT doubling as
"nothing said yet", a real DIRECT server followed by an OWNER was
accepted, and that DIRECT server's own tool calls would afterwards read
OWNER and refuse the logins it is supposed to perform.

The claim refuses a second, different role, and that check lives in the
setter so every entry point is covered by one rule. Two servers in one
interpreter cannot both be right about whether this process may open a
login window, so it is an error to ask for rather than a state to
half-configure. Re-stating the same role is allowed, which is what the
owner does: it claims OWNER at its entry point, ahead of anything that
could fail, and again when it assembles.

No gate reads it yet.

See also: #606

PR #633, opened 2026-07-31
fix(errors): Keep a re-raised tool error's cause

stickerdaniel:

PR #635, opened 2026-07-31
feat(daemon): Keep owner out of the login flow

stickerdaniel:
A detached owner has nobody in front of it. It can open a browser, and
does when configured headed, but an interactive login is different in
kind: it waits for someone to type credentials and answer a challenge,
and no client is attached to this process to do that. It also must not
move the shared profile aside, because the frontend that will log in
needs the session state as the owner saw it.

So the owner reports and the frontend acts. Three gates, keyed on the
role rather than on drives_browser, which is False for a proxy that
legitimately does open a browser for its login:

- the pre-tool gate reports a missing session instead of starting a
  login or an auto-import;
- the stale path reports instead of rotating and relaunching;
- auto-import is refused as a predicate, so gating one call site cannot
  be undone by the next caller.

Closing the browser is not enough by itself. get_or_create_browser opens
one whenever the singleton is None, and readiness asks whether the auth
files exist rather than whether they work, so the next forwarded call
reopened Chromium into the middle of the frontend's login. Measured. A
latch holds until a usable session appears.

Both halves of that condition are needed. A changed generation alone
lifts on an abandoned login, because the frontend rotates first and the
generation then reads as None. _auth_ready() alone never lifts, because
it was already true of the broken session.

An unconfirmed close is reported rather than latched. close_browser keeps
the lease when Chromium's shutdown times out, so inviting a login there
sends the client after a lease it can never take, and the latch would
then answer every later call from a state only a restart clears.
Reproduced before the guard existed.

See also: #606

### DIFF
--- a/linkedin_mcp_server/bootstrap.py
+++ b/linkedin_mcp_server/bootstrap.py
@@ -30,13 +30,17 @@ from linkedin_mcp_server.exceptions import (
     AuthenticationBootstrapFailedError,
     AuthenticationInProgressError,
     AuthenticationStartedError,
+    AuthMissingOnOwnerError,
+    AuthStaleOnOwnerError,
     BrowserSetupFailedError,
     BrowserSetupInProgressError,
     DockerHostLoginRequiredError,
 )
+from linkedin_mcp_server.server_role import ServerRole, process_role
 from linkedin_mcp_server.session_state import (
     auth_root_dir,
     get_runtime_id,
+    load_source_state,
     portable_cookie_path,
     profile_exists,
     rotate_source_profile,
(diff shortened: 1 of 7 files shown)


stickerdaniel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/635?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#637** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/637?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#636** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/636?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#635** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/635?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/635?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#634** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/634?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#633** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/633?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #636, opened 2026-07-31
feat(daemon): Let the client sign in remotely

stickerdaniel:
The owner notices bad LinkedIn auth and cannot repair it; the frontend
was spawned by an MCP client, so it has somewhere to put a login window.
This carries the fact across the loopback hop and acts on it there.

An error result, not an exception. Measured: a middleware that re-raises
delivers meta=None and a generic message to the outer client, because
masking has already discarded the type. The same middleware returning
ToolResult(meta=..., is_error=True) delivers the marker intact, and a
client that raises on error still gets the owner's own wording, because
raise_tool_error has already shaped it into a ToolError.

The reason is read from the exception chain rather than a ContextVar.
Both work today, but fastmcp offloads sync tools to a thread, where a
value set in the body is invisible to the middleware; every tool here is
async, so the ContextVar version would pass and break silently later.

reason and replayable answer different questions and neither implies the
other. reason says what to repair. replayable says whether any work had
happened, so whether the call may run again: a stale session found by the
readiness check has done nothing, while one found mid-scrape may have
sent a message. Only the owner can tell, because all 18 tool catch sites
collapse both origins into one except clause.

A marker whose version or field types this build does not recognise is
ignored rather than half-read. An older frontend attaches to a newer
owner on purpose, so that case is reachable; the reverse is not, because
a newer frontend stands the old owner down first.

See also: #606

PR #637, opened 2026-07-31
feat(daemon): Let one client win the login race

stickerdaniel:
feat(daemon): Let one client win the login race

Two frontends can be told to sign in for the same dead session, and the
lock around the invalidation is process-local: it serializes each process
against itself and knows nothing about the other one.

Rotation itself is safe, because it takes the profile lease, so nobody
can rotate while a login holds it. The exposed window is afterwards: A
finishes a login and writes a generation, B arrives with the old one
still in hand, finds the lease free, and rotates. Measured,
load_source_state then returned None and A's fresh session was gone. The
test reproduces that destruction before asserting the fix.

So the generation the owner found broken travels in the marker, and the
invalidation skips its rotation when the one on disk has moved on. A
generation is only written after a login has validated its session, so a
different one means somebody else's repair succeeded.

Both halves of that check are needed. Without the readiness half, an
abandoned login leaves the profile rotated and the generation reading as
None, which differs from the stale one, so the next client would be told
another client had signed in when the profile is empty and no login would
ever start.

The login also waits for the profile now instead of demanding it. That
path is what a frontend takes when the shared browser asks it to sign in,
and a proxy holds no lease of its own to reuse, so any momentary holder
would leave the user with a login that refuses to open. Bounded by how
long a handover takes rather than how long a person takes, because past
that the holder is a process that will not release.

See also: #606

test(daemon): Prove the marker over a real socket

Everything else about the auth marker runs in memory against a stand-in
owner. This is the one that starts a real detached owner with no LinkedIn
session and asserts the marker arrives over the authenticated loopback
hop as a result rather than being flattened into a generic failure, which
is what an exception would be.

Observed at a plain client with the repair middleware removed, and that
removal is what makes the test possible: create_mcp_server installs it
for every proxy, correctly, so a real frontend consumes the marker, opens
a login window and waits out the login timeout. The test therefore
asserts what the owner sends; the consume-and-replay half stays in memory
with a stubbed login.

The browser cache is symlinked from the account's own. It lives under the
auth root, so an isolated root looks like a fresh install and the
readiness gate reports a download in progress before authentication is
decided at all. Nothing here launches Chromium.

Mutation-checked, including one that survives: removing the role claim in
daemon_owner.main alone changes nothing, because create_mcp_server claims
it again before any call arrives. Removing both fails. The redundancy is
deliberate and this test cannot tell the two apart, which is recorded
next to the assertions.

See also: #606

fix(daemon): Check the generation under the lease

The generation check was in the wrong place. It ran before the login was
even started, and the login then queues for the profile: by the time it
holds one, its opinion can be minutes old. Measured on the real rotation
path, with only the browser stubbed: the winner wrote a fresh generation,
the waiting loser took the lease, rotated, and load_source_state returned
None. The fresh session was gone.

So interactive_login checks again once it holds the profile, which is the
first moment the answer cannot change underneath it. The earlier check
stays: it catches a peer that finished before this client asked at all,
costs nothing, and both are visible in the tests. The missing path is
guarded too, which it was not: a missing session is diagnosed from disk,
so two frontends reach it at once and the login rotates regardless.

The generation now rides on the owner's exception rather than being read
again when the marker is built. The browser has closed by then and the
lease is free, so a second read can return a generation another process
has just written, and the frontend would rotate away the very session it
was told about.

Marker version 2, because the reader from the previous commit accepts a
version-1 marker, ignores the field it does not know, and repairs without
passing the generation on. That is the unguarded path, and an older
frontend attaching to a newer owner is supported on purpose. Refusing the
marker leaves it reporting the owner's failure instead, which is less
informed and honest.

The 60-second wait is a user-experience limit, not a derived worst case:
a holder part way through a tool call does not hand over until it
finishes, and that runs to the tool timeout. The comments said otherwise.

See also: #606

fix(daemon): Guard every path that rotates

The previous guard covered one caller and one shape of the question, and
three ways past it remained.

None meant both "the owner observed no session" and "nothing asked for a
guard", so two clients meeting a genuinely empty profile were unprotected
from each other. Reproduced: the second rotated away the session the
first had just created. A sentinel separates the two, and `--login` typed
at a terminal keeps signing in unconditionally, which is what it is for.

The quiescent gate raised its failures without a generation. Only the
first call reaches the frontend through handle_auth_error; every one
after it is answered by the latch, so the second client to ask got a
marker with nothing to compare against and repaired unguarded.

The auto-import rotates as soon as it holds the profile, exactly as the
login does, and ran ahead of it without the guard. A client whose
keychain read took longer would retire a session a faster peer had
already imported.

The check itself sat outside the lease's try, so a failure while asking
kept the profile until the process exited. An unreadable profile
directory was enough.

The predicate moved to session_state, which owns the generation and is
already imported by both callers.

See also: #606

fix(auth): Let a held profile delay a login

invalidate_auth_and_trigger_relogin retires the dead session before it
starts a login, and that rotation takes the profile without waiting for
it. A holder anywhere else made it raise, and the caller then reported
"the browser profile is in use" without attempting anything.

Refusing there was right when the login also demanded the profile
outright: if the rotation could not take it, neither could the login.
The login waits now, so a momentary holder is exactly the case it was
built for, and failing first throws the wait away before it happens.
Measured with a lease held for 1.5 seconds: the 60-second wait was never
reached.

The rotation is not lost. interactive_login rotates under the profile it
waited for, which is the better place for it anyway, because this one
holds nothing and its result was never guaranteed to survive to the
login. What the failure costs is only that readiness keeps reporting the
dead session until the login retires it, and on a shared owner the
quiescence latch already covers that window.

See also: #606

fix(auth): Pass on the session found wanting

The readiness gate has just looked at the profile when it starts a login,
so it knows which session it found unusable, but it threw that away and
started an unguarded one. Two stdio servers meeting the same bad session
could therefore both sign in, and this is the path every user runs today,
since the shared-browser mode is off by default.

Measured, the damage is bounded: the only case where the earlier session
disappears is one where a second person actually signed in, and then the
later session winning is the expected outcome rather than lost work. What
the guard saves is the pointless second login window.

--login keeps the unguarded default. Somebody typing it is insisting, not
asking, and that is the difference between an override and a request.

Both branches of the gate carry it. A configured Chrome executable skips
the managed-binary path and reaches its own call site; mutating one and
not the other went unnoticed until the test covered both.

See also: #606

fix(auth): Stop a server skipping its own login

A regression from the sentinel work. invalidate_auth_and_trigger_relogin
still used None to mean "no guard requested", and a single-process server
reached it with exactly that, because only the shared owner read the
generation before closing the browser.

The result was the worst available failure. With another process holding
the profile for a moment, the pre-rotation failed and was swallowed, the
login then found the dead session still in place, read it as a peer's
repair, and stood down. The server reported that a login window had
opened. None had. The dead session stayed, and readiness went back to
saying it was fine.

So every role reads the generation now, and the sentinel reaches this
function too, which keeps "observed no session" apart from "nothing asked
for a guard" here as well.

Two tests asserted the old behaviour and were wrong rather than broken:
one required that a single-process server read nothing, which is what
caused this. The new end-to-end test lets the real login run instead of
stubbing the task away, which is what the earlier one missed.

The explicit commands are pinned at their call sites now. Asserting the
helper's default left a mutation passing a generation from --login green
while the real command would have skipped the login it was asked for.

See also: #606

fix(auth): Refuse a login the profile forbids

Two holes the previous round left, both found by review.

An unconfirmed browser teardown keeps the profile lease until the process
exits, and the check for it was owner-only. A single-process server
carried on into a login whose own rotation is refused for exactly that
reason, after having reported that a window had opened. Measured: no
window, the dead session still there, and a state only a restart clears.
The check now runs for every role, ahead of the split, because nothing
that follows can work for either of them.

The generation guard read the generation from the profile it was given
and then asked the *configured* profile whether a session was usable.
Both callers accept an explicit directory, so the two can differ, and
then a peer's fresh session was rotated away because a different profile
happened to be empty. Readiness takes the directory now.

That also removed a redundant hop: readiness went through a helper that
resolves the profile itself and re-checks two things the caller has
already tested. It now reads the state file directly, which is what makes
it answerable for a directory handed in.

The DIRECT test asserted only that the keyword was present, so a mutation
passing a hardcoded None stayed green. It asserts the value now.

See also: #606

fix(auth): Compare generations under the lock

Every automatic path decided whether to retire a session and then called a
rotation that takes the profile itself. A peer completing a login between
those two points had its fresh session quarantined by a process acting on
a view formed before that session existed.

Five rounds of review missed this, including one of my own attempts:
reproducing it needs the peer to land at the rotation boundary rather
than anywhere earlier, and the earlier guards were close enough to the
decision to look like they covered it.

So the comparison moves into rotate_source_profile, inside the lock it
already takes. There the answer cannot change between reading it and
acting on it, which is the property the call sites could not offer at any
distance. Both automatic callers hand their observed generation through,
and a test pins each, because the comparison is only under the lock if
the generation reaches it.

The early guard in invalidate_auth_and_trigger_relogin is gone. It is
redundant now, and its outcome was worse: it refused the whole repair
where the rotation simply retires nothing and lets the login proceed. A
client asking to repair a session should not be left with nothing because
somebody else got there first.

The sentinel and the predicate move to session_state with it. That module
owns the generation, and the rotation cannot import them from setup.

_has_source_state is deleted. The last commit removed its only caller and
left a docstring claiming otherwise.

See also: #606

fix(auth): Report a peer win as what it is

The rotation signalled "I preserved a peer's session" the same way it
signals "there was nothing to retire": by returning None. The caller
could not tell them apart, so it carried on, scheduled a login and told
the user a browser window had opened. The task then took the profile, saw
the peer's session, and stood down without opening anything.

The data was safe. The report was not, and it is the same user-visible
lie as an earlier round's blocker, without the wedge behind it. The test
I had changed to accept AuthenticationStartedError there blessed exactly
that outcome, and did not await the task, so nothing noticed.

A distinct exception now carries the outcome, with its own diagnostics-
free branch in the error handler: nothing failed, somebody else simply
got there first, and the honest answer is to retry and use their session.

See also: #606

fix(daemon): Replay after a peer signs in

The frontend caught the peer-win outcome with its general failure handler,
so a repair that had in fact succeeded, by somebody else, refused the
replay. The user was sent back for a retry while the session the call
needed was already on disk.

The two side effects reported alongside it are already gone with the
previous commit: no login task is scheduled, so nothing reaches
_ensure_full_chromium_installed, and no headed browser is downloaded for
a window that was never going to open. Verified on both the stale and the
missing path.

See also: #606

fix(daemon): Wait out the sign-in before replaying

A started login reports itself by raising, while the window is still open
and nobody has typed anything yet. The frontend read that as a failed
repair and refused the replay, so every repair this middleware exists to
perform was abandoned one step before it worked.

Measured on the stale path with a login that finished successfully: the
owner was called once and the client still got the error. The missing
path did the same as soon as signing in took longer than the inline
budget, which is every login a person actually performs.

The tests missed it because they replaced the repair with a mock that
returned on command, which is exactly the step that raises.

See also: #606

fix(daemon): Let a wedged owner give way

An unconfirmed browser shutdown keeps the profile lease until the process
exits. For a single-process server the resulting advice is actionable:
its client restarts it. A detached owner outlives that client, so the
next frontend attaches straight back to it and meets the same held
profile, and only killing it by hand recovered.

It now asks to stand down. The serve loop finishes what is in flight and
exits, the kernel frees the daemon lock, and the next call elects a fresh
owner. Nothing is lost, because the session lives on disk.

The frontend login has the same unconfirmed-shutdown hole and is left
alone deliberately: there a client restart does recover.

See also: #606

fix(daemon): Keep the sign-in wait inside its call

The wait was a fixed 150 seconds, chosen against the 180-second default
tool timeout. TOOL_TIMEOUT is a user setting: measured at 60, the
client's call was cut short while the middleware was still waiting, so
instead of the readable 'sign-in still in progress' the user got a bare
timeout. The login survived and the next call worked, so nothing was
lost but the explanation, which is the part a person acts on.

It is now a fraction of the configured budget, minus what the call has
already spent. Subtracting matters: the fraction alone still overran,
because the owner call and the inline login wait come out of the same
budget before this is reached.

See also: #606

fix(daemon): Decide the stand-down in a finally

close_browser defers cancellation until teardown finishes and then
re-raises CancelledError, which is a BaseException and passes straight
through the except Exception around it. A tool timeout landing during
the close therefore skipped the whole stand-down decision, and a wedged
owner stayed wedged: recoverable only by killing it.

Measured: wait_for(handle_auth_error(...), 0.02) against a close that
defers its cancel left stand_down_reason() at None with the lease held.

The decision now runs in a finally, so it survives the cancellation.
Raising the error stays on the normal path, because a cancelled call has
nobody left to raise it to.

See also: #606

fix(daemon): Bound the replay by what is left

Nothing under this middleware bounds the whole path. The forwarded call
carries its own deadline per hop, so the first call, the wait and the
replay each stay inside one while their sum stays inside nothing. A
sign-in finishing near the end of the wait, followed by a slow replay,
put the client past its deadline with the answer already in hand.

The replay now gets what the call has not spent, with a floor so an
already-overrun call still makes a real attempt. When it runs out, the
owner's readable failure is the answer: the session exists by then, so
the next call succeeds, and a bare transport error would say none of
that.

See also: #606

fix(daemon): Budget the repair from its own server

The repair read tool_timeout_seconds from the configuration singleton
while create_mcp_server is handed the budget every tool it registers
gets. The two agree for a server cli_main built, which loaded that
configuration first, and need not for one constructed directly:
measured at tool_timeout=1.2 with no configuration loaded, the repair
budgeted itself 149.8 seconds of a 1.2-second call, because the unloaded
singleton falls back to parsing sys.argv.

The middleware now takes the budget it was built with.

The budget test also asserted only 'less than the timeout', which left
the subtraction untested: dropping it yields five sixths of the budget,
still under, and the overrun the test exists for came back green.

See also: #606

fix(daemon): Never abandon a mutating replay

Timing the replay out does not stop it. Cancellation is not forwarded
across the hop, which daemon_proxy already records, so the frontend
reported the old auth failure while the owner went on to send the
message. Measured over a real loopback owner: the client was answered at
0.66s with an error and the owner appended its effect 0.7s later. The
user is told the message was not sent, it was, and retrying sends it
twice.

The bound now applies only to tools that declare readOnlyHint, where an
abandoned replay costs a wasted page load. Anything else, including a
tool that declares nothing, runs to completion: an overrun call is the
client's own deadline to give up on, and that leaves one call
outstanding rather than a second queued behind it.

The in-memory test could not see this. ProxyClient delivers cancellation
into the owner's coroutine, so it proved a property the HTTP path does
not have.

Also caps the replay floor by the budget: at tool_timeout_seconds=1.0
with 0.9s spent it allowed 5 seconds.

See also: #606

fix(daemon): Leave a changing call for the caller to retry

No bound on an automatic replay can make it safe. Bounding it locally
left the owner finishing the mutation after the frontend had reported
failure, because cancellation is not forwarded across the hop. Removing
the bound only moved the deadline: an MCP client cancelling its own call
produced the same orphan, and the retry sent the message twice.

Nothing this process can do closes that, because the deadline that
matters belongs to a client it does not control. So the auth is repaired
and the owner's own message, which already says to retry, is returned.
The user knows whether it happened; asking costs one retry, guessing
wrong sends it twice. Read-only tools keep the replay.

get_conversation and search_conversations lose readOnlyHint. Both
enumerate by visiting rows, which marks conversations as read, and both
docstrings always said so. An unread message is state.

See also: #606

test(daemon): Pin the budget arithmetic off the clock

The budget test subtracted a duration measured from the start of the
test, which includes client setup the middleware's own clock never saw.
It passed alone and failed inside the full suite at 1.22s against 1.2s.
That measured the runner.

It now asserts the timeout asked for against the share, plus the
subtraction directly, so the property is pinned rather than inferred
from wall clock.

See also: #606

fix(daemon): Free an owner stranded by any close

Three ways into one dead end, found one at a time, so the decision now
lives in one place that every one of them reaches.

The routine close is the path that hid longest: a handoff, the idle
timeout and close_session all return normally when the teardown cannot
be confirmed. They log, keep the lease, and raise nothing, so neither
the exception nor handle_auth_error ever runs. Reproduced: close
returned, lease kept, stand_down_reason None, next launch refused.

The cookie export could strand the profile even earlier. It awaits a
protocol call with no deadline, before the bounded teardown, with the
singleton already cleared and cancellation deferred. It is now bounded
like the teardown steps below it; a lost export costs the Docker cookie
file this run, which the next close rewrites.

See also: #606

fix(daemon): Ask after the profile is marked held

The startup path was the one measured live, and it was still open. The
exception's constructor runs while _create_browser_locked is raising,
when browser_open is false, so the helper reads that nothing is held and
returns. Only the catch afterwards marks the browser open and takes the
extra reference, holding the profile with nobody having asked for a
replacement.

Reproduced in production order: browser_open=True, stand_down=None.

The test for this handed the constructor a lease that already said True,
which is the state production has not reached yet. That mock is what hid
it, and the comment above it asserted the opposite of what the code
does. It now drives _create_browser and lets the marker start false.

See also: #606

fix(daemon): Treat an unreadable lease as held

Not being able to tell whether the profile is held is not the same as
being told it is free, and the two outcomes are not symmetric: an
unnecessary stand-down costs one election, a stranded owner costs every
later call until somebody kills it by hand. Measured with the read
raising PermissionError: the owner kept serving with the question
unanswered.

The guard test that was supposed to keep this from firing on healthy
owners drove a confirmed close, which never reaches the helper at all,
so it stayed green with the guard deleted. It now asserts the helper
directly and fails on that mutation.

Also drops the raise-site count from the docstring. It said seven, then
nine; there are eight, and the ninth match was the class itself.

See also: #606

fix(daemon): Ask when the close cannot read the lease

The lookup sits between the teardown and every line that protects the
profile. It resolves a path, so it can fail on a profile directory that
became unreadable while the browser was running, even though this
process still holds the open descriptor. By then the singleton is
cleared, so nothing below it runs: not the hold flag, not the warning,
not the request.

Measured in production order with an unconfirmed close and the lookup
raising: browser cleared, hold flag still set, no stand-down.

Asked before re-raising, and only for an unconfirmed teardown. A proven
close means the profile is free, so a failed lookup there is the
caller's problem rather than a reason to end the owner. The original
error reaches the caller either way.

See also: #606

fix(daemon): Ask on any failed lease lookup at close

The confirmed case was exempted on the reasoning that a proven-closed
Chromium leaves the profile free. It does not: mark_browser_closed() and
release() both need the object the lookup failed to produce, so a
confirmed close strands the lease just as thoroughly.

Measured with a real acquired lease: Chromium gone, the lease still held
with its marker set, and this owner's own next launch refused with
BrowserBusyError while nobody had asked for a replacement.

The test for that branch asserted the exemption, so it pinned the wrong
recovery. It now asserts the request, and fails when the exemption is
restored.

See also: #606

### DIFF
--- a/linkedin_mcp_server/bootstrap.py
+++ b/linkedin_mcp_server/bootstrap.py
@@ -16,7 +16,6 @@ from typing import NoReturn
 
 from fastmcp import Context
 
-from linkedin_mcp_server.authentication import get_authentication_source
 from linkedin_mcp_server.common_utils import secure_mkdir, secure_write_text, utcnow_iso
 from linkedin_mcp_server.config import get_config
 from linkedin_mcp_server.config.schema import is_loopback_host
@@ -663,20 +662,6 @@ def _auth_ready(profile_dir: Path | None = None) -> bool:
     )
 
 
-def _has_source_state() -> bool:
-    """Whether the configured profile has a readable source state.
-
-    Kept because other callers use it. Readiness no longer goes through it: it
-    resolves the profile itself, so it could not answer for a directory handed in
-    explicitly, and it re-checks the two things the caller has already tested.
-    """
-    try:
-        get_authentication_source()
-    except Exception:
-        return False
-    return True
-
-
 def _auto_import_allowed() -> bool:
     """Return whether a silent browser-session import is safe to attempt now.
 
@@ -960,7 +945,7 @@ async def _start_login_if_needed(
                 prior_error = None
             else:
                 prior_error = _state.last_error
-                _move_invalid_auth_state_aside()
+                _move_invalid_auth_state_aside(superseded_by)
                 _state.auth_state = AuthState.STARTING
                 _state.auth_started_at = utcnow_iso()
                 _state.last_error = None
@@ -1148,44 +1133,6 @@ async def invalidate_auth_and_trigger_relogin(
                 "then retry this tool."
             )
 
-        # Checked here, after the lock and immediately before the rotation, which
-        # is the only place it means anything. Two frontends can meet the same
-        # dead session, and `_lock` is process-local: it serializes this function
-        # against itself in *this* interpreter and knows nothing about the other
-        # process.
-        #
-        # Measured. Rotation itself is safe, because it takes the profile lease
-        # (`session_state.py:727`), so nobody can rotate while a login holds it.
-        # The exposed window is afterwards: A finishes a login and writes a
-        # generation, B arrives with the *old* one still in hand, finds the lease
-        # free, and rotates. `load_source_state` then returned None, and A's fresh
-        # session was gone.
-        #
-        # Comparing generations closes it, because one is only written by a login
-        # or import that has already validated its session. A generation different
-        # from the one being complained about therefore means somebody else's
-        # repair succeeded, and the right move is to stop.
-        #
-        # This is the early half of a pair, and it is not the important one. It
-        # runs before any waiting, so it only catches a peer that finished before
-        # this client even asked. The window that actually loses data opens after
-        # this returns: the login it starts queues for the profile, and by the
-        # time it holds one its opinion may be stale. `interactive_login` checks
-        # again there, where the answer can no longer change. Removing either is
-        # visible in the tests, so neither is decoration.
-        if stale_generation is not UNGUARDED:
-            on_disk = current_login_generation()
-            if on_disk != stale_generation and _auth_ready():
-                logger.info(
-                    "Another client already signed in (generation %s); leaving "
-                    "its session alone",
-                    on_disk,
-                )
-                raise AuthenticationBootstrapFailedError(
-                    "Another LinkedIn MCP client has already signed in. Retry "
-                    "this tool to use the new session."
-                )
-
         # Force-move stale profile files (skip _auth_ready() guard).
         #
         # A failure here used to stop the login, and that reasoning has expired.
@@ -1209,7 +1156,7 @@ async def invalidate_auth_and_trigger_relogin(
         # profile. What must not happen is the login standing down there, which
         # is why the generation it was told about travels with it.
         try:
-            _force_move_auth_state_aside()
+            _force_move_auth_state_aside(stale_generation)
         except AuthenticationBootstrapFailedError as exc:
             logger.info(
                 "Could not retire the stale session yet (%s); the login will "
@@ -1245,7 +1192,9 @@ async def invalidate_auth_and_trigger_relogin(
     )
 
 
-def _move_auth_state_aside(*, force: bool = False) -> None:
+def _move_auth_state_aside(
+    *, force: bool = False, superseded_by: str | None | object = UNGUARDED
+) -> None:
     """Move auth artifacts to a timestamped backup directory.
 
     Args:
@@ -1267,7 +1216,7 @@ def _move_auth_state_aside(*, force: bool = False) -> None:
     # Quarantine creation lives in session_state so the routine rotation on a
     # new session and this stale-state path produce identically shaped backups.
     try:
-        rotate_source_profile(get_profile_dir())
+        rotate_source_profile(get_profile_dir(), superseded_by=superseded_by)
     except RuntimeError as exc:
         raise AuthenticationBootstrapFailedError(
             f"{exc} No login was started."
@@ -1278,13 +1227,17 @@ def _move_auth_state_aside(*, force: bool = False) -> None:
         ) from exc
 
 
-def _force_move_auth_state_aside() -> None:
+def _force_move_auth_state_aside(
+    superseded_by: str | None | object = UNGUARDED,
+) -> None:
     """Move auth artifacts aside unconditionally (no ``_auth_ready()`` guard)."""
-    _move_auth_state_aside(force=True)
+    _move_auth_state_aside(force=True, superseded_by=superseded_by)
 
 
-def _move_invalid_auth_state_aside() -> None:
-    _move_auth_state_aside(force=False)
+def _move_invalid_auth_state_aside(
+    superseded_by: str | None | object = UNGUARDED,
+) -> None:
+    _move_auth_state_aside(force=False, superseded_by=superseded_by)
 
 
 async def _run_login_flow() -> None:
--- a/linkedin_mcp_server/bootstrap.py
+++ b/linkedin_mcp_server/bootstrap.py
@@ -46,7 +46,7 @@ from linkedin_mcp_server.session_state import (
     rotate_source_profile,
     source_state_path,
 )
-from linkedin_mcp_server.setup import interactive_login
+from linkedin_mcp_server.setup import UNGUARDED, interactive_login
 
 logger = logging.getLogger(__name__)
 
@@ -113,7 +113,7 @@ class BootstrapState:
     #: signature: `_run_login_flow` is spawned as a bare task in several places
     #: and stubbed in as many tests, and threading an argument through all of
     #: them would change far more than the one thing that matters.
-    login_supersedes: str | None = None
+    login_supersedes: str | None | object = UNGUARDED
 
 
 _state = BootstrapState()
@@ -138,6 +138,7 @@ def reset_bootstrap_for_testing() -> None:
     _state = BootstrapState()
     _lock = asyncio.Lock()
     _AUTO_IMPORT_ANNOUNCED = False
+    _state.login_supersedes = UNGUARDED
     _auth_quiescent = False
     _auth_quiescent_generation = None
     os.environ.pop("PLAYWRIGHT_BROWSERS_PATH", None)
@@ -801,7 +802,15 @@ async def _try_auto_import_session(ctx: Context | None = None) -> bool:
         # reads are already bounded (security 10s / secret-tool 10s); this covers
         # the launch + navigation budget on top.
         result = await asyncio.wait_for(
-            import_session_from_browser(None, user_data_dir=user_data_dir),
+            import_session_from_browser(
+                None,
+                user_data_dir=user_data_dir,
+                # The import rotates the profile as soon as it holds the lease,
+                # exactly as the login does, so it needs the same guard. Without
+                # it a client whose keychain read ran long would retire a session
+                # a faster peer had already imported.
+                superseded_by=_state.login_supersedes,
+            ),
             timeout=60,
         )
         if not result:
@@ -835,7 +844,7 @@ async def _try_auto_import_session(ctx: Context | None = None) -> bool:
 
 
 async def _start_login_if_needed(
-    ctx: Context | None = None, *, superseded_by: str | None = None
+    ctx: Context | None = None, *, superseded_by: str | None | object = UNGUARDED
 ) -> None:
     if process_role() is ServerRole.OWNER:
         # Ahead of the lock, and ahead of every branch below, because all of them
@@ -880,6 +889,7 @@ async def _start_login_if_needed(
             # Claim the one-shot import under the lock so only one keychain read
             # / import browser ever runs per process episode.
             _state.import_attempted = True
+            _state.login_supersedes = superseded_by
             _state.import_task = asyncio.create_task(
                 _try_auto_import_session(ctx), name="linkedin-auto-import"
             )
@@ -968,7 +978,7 @@ async def _start_login_if_needed(
 
 
 async def start_login_if_needed(
-    ctx: Context | None = None, *, superseded_by: str | None = None
+    ctx: Context | None = None, *, superseded_by: str | None | object = UNGUARDED
 ) -> None:
     """Public wrapper for starting the shared login workflow."""
     await _start_login_if_needed(ctx, superseded_by=superseded_by)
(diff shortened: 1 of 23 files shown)


stickerdaniel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/637?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#637** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/637?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/637?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#636** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/636?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#635** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/635?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#634** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/634?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#633** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/633?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #645, opened 2026-07-31
fix(daemon): Ask the lease the browser actually holds

### LINKED ISSUE #642: fix(daemon): The handoff asks a lease the browser does not hold
# The handoff asks a lease the browser does not hold

**Setup:** internal, `daemon_enabled` still off. No user-visible symptom today.

## What Happened

`release_profile_if_idle_or_requested()` (`drivers/browser.py`) null-checks the retained `_browser_lease`, then twenty lines later calls `get_profile_lease()` and asks *that* object whether a process is waiting and how long the profile has been held.

Both answers belong to the lease this browser took:

- `handoff_requested()` probes a file under the lease's own auth root, so a resolved one answers about a profile nobody is waiting for. A real waiter is ignored, or a stale signal hands over a browser nobody asked for.
- `held_seconds` counts from acquisition, so a freshly resolved object reports `0.0`, the hold window never elapses, and the handover is refused forever. That is the one that would strand a waiter outright.

## Cause

#641 removed this reconstruction from the close path and from the stand-down decision. It left the one in the handoff decision standing.

## Scope

Latent rather than live, and worth saying plainly because a reader looking for a user-visible symptom will not find one. Under the shipped CLI the registry is keyed by the resolved auth root and nothing changes `user_data_dir` while a browser is open, so the two objects are the same. Where they can diverge: an embedder or test that changes the configured profile mid-session, a retargeted symlink under the profile path, a cleared registry.

It is the shape #641 exists to remove: ownership settled through a value that was looked up rather than held.

## Notes

- The three tests covering this function each patch the resolver to return the same object they inject as `_browser_lease`, so none of them can tell which one the code consulted. A mutation reverting the fix passes all three.
- Follow-up worth recording rather than fixing here: on re-entry after an unconfirmed close, `_create_browser` also resolves a lease while already holding one, before the `browser_open` refusal.

See also: #606

stickerdaniel:
Closes #642

`release_profile_if_idle_or_requested()` null-checks the lease the browser holds and then, twenty lines later, throws that object away and asks a freshly resolved one whether a process is waiting and how long the profile has been held. Both questions are about *this* profile: the handoff probe reads a file under the lease's own auth root, and the hold window counts from acquisition, so a resolved object reports a hold of zero and the window never elapses. That second one is what would strand a waiter outright.

#641 removed this reconstruction from the close path and from the stand-down decision. This is the one it left standing, and it is the same shape: ownership settled through a value that was looked up rather than held.

Worth saying plainly, because a reader expecting a user-visible symptom will not find one. Under the shipped CLI the two objects are the same: the registry caches one lease per resolved auth root, the configuration is loaded once, and no shipped path changes `user_data_dir` or resets the config singleton while a browser is open. This is latent.

The part worth a look in review is the tests. All three covering this function patched the resolver to return the same object they injected as `_browser_lease`, so none of them could tell which one the code consulted, and a mutation reverting the fix passed all three. Those patches are gone rather than tidied away: leaving them would let a reintroduced lookup hide behind them exactly as this one did. Two new tests hand out different objects, which is the only way anything here can see the difference.

Verified with the full suite, lint, format and type checks. Both new tests were mutation-checked: reverting `lease = _browser_lease` fails the first on the resolved lease reporting no waiter, and the second on it reporting a hold of zero, while the three existing tests stay green.

## Synthetic prompt

> `release_profile_if_idle_or_requested()` in `drivers/browser.py` still calls `get_profile_lease()` and asks that object `handoff_requested()` and `held_seconds`, even though #641 made the browser retain its own `ProfileLease`. Use the retained one. Then check whether the existing tests could ever have caught this, and fix them if not.

### DIFF
--- a/linkedin_mcp_server/drivers/browser.py
+++ b/linkedin_mcp_server/drivers/browser.py
@@ -992,7 +992,16 @@ async def release_profile_if_idle_or_requested() -> bool:
         return False
 
     config = get_config().browser
-    lease = get_profile_lease()
+    # The lease this browser holds, not one the registry resolves now. Both
+    # questions asked below are about *this* profile: `handoff_requested()`
+    # probes a file under the lease's own auth root, and `held_seconds` counts
+    # from the moment that lease was acquired. A freshly resolved object answers
+    # about whichever path the config points at now and reports a hold of zero,
+    # so a waiter would be either ignored or refused forever. `config` still
+    # comes from settings: the hold window and the idle timeout are not lease
+    # state. Bound to a local so this stays true if an `await` ever appears
+    # above it and lets a concurrent close clear the global.
+    lease = _browser_lease
     # None means no tool call has ever run, which is idle in the strongest sense.
     idle_for = time.monotonic() - _last_activity if _last_activity is not None else None
 
--- a/tests/test_profile_lease_integration.py
+++ b/tests/test_profile_lease_integration.py
@@ -295,14 +295,11 @@ class TestIdleOwnerHandsOver:
         fake_browser = MagicMock()
         fake_browser.close = AsyncMock(return_value=True)
 
-        with (
-            patch.multiple(
-                browser_module,
-                _browser=fake_browser,
-                _browser_cookie_export_path=None,
-                _browser_lease=lease,
-            ),
-            patch.object(browser_module, "get_profile_lease", return_value=lease),
+        with patch.multiple(
+            browser_module,
+            _browser=fake_browser,
+            _browser_cookie_export_path=None,
+            _browser_lease=lease,
         ):
             watcher = asyncio.create_task(browser_module.watch_for_handoff_requests())
             waiter = subprocess.Popen(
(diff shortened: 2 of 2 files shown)


stickerdaniel:
* **#646** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/646?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#645** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/645?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/645?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
### Merge activity

* **Jul 31, 10:05 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/645).
* **Jul 31, 10:05 PM UTC**: @stickerdaniel merged this pull request with [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/645).

PR #646, opened 2026-07-31
fix(daemon): Re-check for work before handing the browser over

### LINKED ISSUE #643: fix(daemon): A tool call can lose its browser mid-handover
# A tool call can lose its browser mid-handover

**Setup:** internal, `daemon_enabled` still off. Reachable in production once it is on.

## What Happened

`release_profile_if_idle_or_requested()` refuses to close a browser a tool call is driving:

```python
if _calls_in_flight > 0:
    return False
```

The check and the teardown are not adjacent. `close_browser()` first waits for `_browser_lifecycle_lock`, which a close already in progress holds across the cookie export and the Chromium shutdown. A call that starts inside that wait is invisible to a check that already ran.

Measured, with the lock held to open the window: the decision was taken with nothing in flight, and Chromium was closed and the profile released with one call running. That call's `Page` is gone, which is exactly the closed-target error the guard exists to prevent.

## Reachability

The background poller runs for every server (`server.py:102`) and `note_call_started()` is on the ordinary tool path (`sequential_tool_middleware.py:120`). It needs a call to arrive inside the lock wait, so it is rare rather than impossible; a slow or concurrent teardown widens the window.

## Proposal

Re-establish the condition before the teardown proceeds, in the caller whose close is *conditional*.

Not inside `close_browser()`. That is the general teardown entry point, and `close_session`, the shutdown path, login and browser import all mean "close it" unconditionally. A counter check there would make a deliberate close silently do nothing.

Placement is the substance rather than a detail: reading the counter and then starting the teardown as a task leaves a scheduling point between the two, because `asyncio.create_task` does not begin the coroutine. The check has to happen in the same executed step as the synchronous detach of `_browser`.

## Notes

The existing test covers the simple case where a call is already in flight at decision time. It cannot see either window, so it stays as it is and the new cases are asserted separately.

See also: #606

stickerdaniel:
Closes #643

Stacked on #645, which touches the neighbouring lines. Read this diff against that one.

The in-flight guard runs, and the close it decides on then waits for the lifecycle lock, which a teardown already in progress holds across the cookie export and the Chromium shutdown. A tool call that starts inside that wait was never seen. Measured with the lock held to open the window: the decision was taken with nothing in flight, and Chromium was closed and the profile released with one call running, whose `Page` is gone. That is exactly the closed-target error the guard's comment exists to prevent.

Unlike #645 this one is reachable. The poller runs for every server and `note_call_started()` is on the ordinary tool path; it needs a call to arrive inside the lock wait, so rare rather than impossible, and any slow or concurrent teardown widens it.

The part worth review is *where* the re-check sits, because the obvious placement is wrong in a way that still passes a test written for it. Reading the counter in the caller and then starting the teardown as a task leaves a scheduling point between the two: `asyncio.create_task` does not begin the coroutine, so anything already runnable goes first. I measured that too, after a review flagged it: the re-check saw zero, a tool call ran next and took the live browser through the fast path in `get_or_create_browser`, and the teardown closed that same browser with one call in flight. So the check lives in the coroutine the task runs, where the body reaches its first `await` without yielding and `_close_browser_locked` clears `_browser` before its own first await.

Not in `close_browser()`: that is the general teardown entry point, and `close_session`, shutdown, login and import all mean "close it" unconditionally. A counter check there would make a deliberate close silently do nothing.

Two tests, because one window hides the other. The first starts its call while the close waits for the lock; the second one scheduling point later. The second fails when the check moves back into the caller while the first two stay green, which is how the narrower window went unnoticed in the first place. Both bring their own lifecycle lock, since `asyncio.Lock` binds to the first loop that awaits it and the module-level one outlives each test.

Verified with the full suite (1584 passed, 10 skipped), lint, format and type checks, plus the mutation runs described above.

## Synthetic prompt

> The `_calls_in_flight` guard in `release_profile_if_idle_or_requested()` is a check-then-act race: `close_browser()` waits on `_browser_lifecycle_lock` first, so a tool call starting during that wait gets its browser torn down. Re-establish the condition before the teardown, in this caller rather than in `close_browser`, and make sure the check and the detach of `_browser` happen in the same executed step. Cover both windows with tests.

### DIFF
--- a/linkedin_mcp_server/drivers/browser.py
+++ b/linkedin_mcp_server/drivers/browser.py
@@ -11,7 +11,8 @@ import logging
 import os
 import time
 from pathlib import Path
-from typing import Any
+from collections.abc import Coroutine
+from typing import Any, TypeVar
 
 from linkedin_mcp_server.common_utils import harden_linkedin_tree, secure_mkdir
 from linkedin_mcp_server.core import (
@@ -93,6 +94,8 @@ _calls_in_flight: int = 0
 # arriving in that window would see no browser and launch a second Chromium on
 # the same profile, which is the very corruption this module prevents.
 _browser_lifecycle_lock = asyncio.Lock()
+
+T = TypeVar("T")
 
 
 def _debug_skip_checkpoint_restart() -> bool:
@@ -773,16 +776,27 @@ async def close_browser() -> None:
     exactly when a new launch must not start.
     """
     async with _browser_lifecycle_lock:
-        task = asyncio.create_task(_close_browser_locked())
-        cancelled = False
-        while True:
-            try:
-                await asyncio.shield(task)
-                break
-            except asyncio.CancelledError:
-                cancelled = True
-        if cancelled:
-            raise asyncio.CancelledError
+        await _run_deferring_cancels(_close_browser_locked())
+
+
+async def _run_deferring_cancels(coroutine: Coroutine[Any, Any, T]) -> T:
+    """Run *coroutine* to completion, holding cancellation back until it ends.
+
+    The caller holds the lifecycle lock. Splitting this out of ``close_browser``
+    lets the conditional close run its own coroutine under the same protection,
+    rather than duplicating the shield loop.
+    """
+    task = asyncio.create_task(coroutine)
+    cancelled = False
+    while True:
+        try:
+            result = await asyncio.shield(task)
+            break
+        except asyncio.CancelledError:
+            cancelled = True
+    if cancelled:
+        raise asyncio.CancelledError
+    return result
 
 
 def _settle_the_profile(*, confirmed: bool) -> None:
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#646** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/646?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/646?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#645** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/645?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

stickerdaniel:
### Merge activity

* **Jul 31, 10:05 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/646).
* **Jul 31, 10:06 PM UTC**: [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/646) rebased this pull request as part of a merge.
* **Jul 31, 10:08 PM UTC**: @stickerdaniel merged this pull request with [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/646).

PR #648, opened 2026-08-01
refactor(daemon): Give the proxy a backend it can re-resolve

stickerdaniel:
A proxy resolves its owner's address and token once, at startup, and nothing
downstream can change them. `create_proxy_provider` receives an `Attachment` and
a timeout, and `create_mcp_server` has no configuration parameter at all, so the
proxy layer could not elect a replacement owner even if it wanted to.

This replaces the bare attachment with a `DaemonProxyBackend` that holds the
current attachment alongside the inputs an election needs, and turns the client
factory into a bound method that reads it per operation. `ProxyProvider` calls
its factory for every upstream operation, so following a replacement needs no
further machinery once the address is read rather than captured. Reaching for
`get_config()` in the proxy layer instead would walk into the argv sensitivity
`daemon_auth` already documents: the first read of an unloaded singleton parses
whatever `sys.argv` holds, which for a directly constructed server is pytest's
command line.

Behaviour is unchanged, and no test's assertion changes. The call sites move with
the signature, two assertions reach the attachment through the backend, and one
helper stopped being a `MagicMock` because the dead-owner test needs production's
own client builder to fail against a dead port.

Nothing replaces the attachment yet. That is the next change in the stack.

The full suite, ruff and ty pass.

See also: #606

## Synthetic prompt

> Give the daemon proxy a state object it can re-resolve. Replace the
> `Attachment` that `create_proxy_provider` and `create_mcp_server` take with a
> `DaemonProxyBackend` holding the attachment plus the auth root, profile and
> config an election would need, make the client factory a bound method that
> reads the current attachment per upstream operation, and carry the inputs
> through `cli_main`. Pure refactor, no behaviour change.

### DIFF
--- a/linkedin_mcp_server/cli_main.py
+++ b/linkedin_mcp_server/cli_main.py
@@ -38,7 +38,7 @@ from linkedin_mcp_server.server import ServerRole, create_mcp_server
 from linkedin_mcp_server.setup import run_profile_creation
 
 if TYPE_CHECKING:
-    from linkedin_mcp_server.daemon import Attachment
+    from linkedin_mcp_server.daemon_proxy import DaemonProxyBackend
 
 logger = logging.getLogger(__name__)
 
@@ -286,7 +286,7 @@ def profile_info_and_exit() -> None:
     sys.exit(1)
 
 
-def _obtain_shared_owner(config: AppConfig) -> "Attachment | None":
+def _obtain_shared_owner(config: AppConfig) -> "DaemonProxyBackend | None":
     """Return the shared owner to forward to, starting one if nobody has.
 
     Off by default, and inert when off. ``None`` means this process serves its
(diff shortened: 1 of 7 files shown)


stickerdaniel:
* **#649** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/649?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#648** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/648?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/648?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
### Merge activity

* **Aug 1, 1:31 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/648).
* **Aug 1, 1:31 PM UTC**: @stickerdaniel merged this pull request with [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/648).

PR #649, opened 2026-08-01
fix(daemon): Recover a proxy whose owner has gone away

stickerdaniel:
A proxy whose owner is replaced failed every operation afterwards. An upgrade is
enough to cause it: `@latest` is the documented install, so the first client
launched after one stands the running owner down, and every proxy already
attached to it stayed dead until its own process restarted. Reproduced end to
end before this change: a proxy serving 19 tools had its owner told to stand down
through the production path, and its next listing failed with
`McpError: Client failed to connect`. A crashed owner and an idle exit end the
same way.

A transport failure now says which owner it was against and whether the request
had left this process, and the backend elects a replacement. The connect is one
boundary and every listing is another: `ProxyProvider` runs its listings inside
a connection that was already established, so an owner dying between the
initialize and the list request raises out of neither `__aenter__` nor
`call_tool_mcp`, and a client that lists before it calls could never reach the
recovery.

A JSON-RPC error is not a departure, because a process that answers is
reachable. Three codes are the exception, and they are what a departure *during*
a request looks like: the client invents `REQUEST_TIMEOUT` when nothing came
back, `CONNECTION_CLOSED` when the read stream closes with requests still
pending, and `32600` when a POST is answered 404 for a session that is gone.
Measured against a real owner killed mid-call: `Timed out while waiting for
response to CallToolRequest`. Nobody sent that, and reading the type rather than
the code would leave the frontend sitting on a process that is not there.
Reading the code is a deployment convention rather than a protocol guarantee,
and it is written down as one: it holds because the owner is this same package,
which constructs no JSON-RPC error anywhere, and it is wrong in the direction
that self-corrects. A spurious election probes, finds the same owner answering
and keeps the attachment, while matching the SDK's message text as well would
mean a reworded message in a future release silently strands the proxy again.

Whether a failed call may be run again has two halves, and both are load-bearing.
The tool's own annotation says whether a repeat could change anything on
LinkedIn, and the failure says whether the request had left this process. There
is no dispatch acknowledgement anywhere in the protocol, so a mutating call is
repeated only when nothing was sent; otherwise the replacement is adopted for the
next call and the original failure is reported. The annotation rule is the one
`daemon_auth` already uses for its sign-in replay, promoted to a public name so
the two cannot drift apart in the unsafe direction.

The election runs in a thread for correctness rather than latency: `obtain_owner`
probes liveness through `asyncio.run`, which raises inside a running loop and is
caught as a refusal, so an in-loop call would bury a healthy owner. One election
runs at a time, keyed by the owner that failed and shielded, because a thread
outlives the cancellation of whoever awaited it.

The provider now keeps no component cache. Clearing it when a replacement is
adopted was the first design and does not hold: `ProxyProvider` writes each cache
after its listing completes and outside any lock, so a listing already in flight
against the departing owner refills a cache that was cleared while it ran, which
is reproducible by holding one open across an adoption. The cost is 27.8ms
against 56.0ms median per forwarded call, measured against a real loopback owner,
for a call that drives a browser through a LinkedIn page for seconds. What it
buys is that a five minute old `readOnlyHint` can no longer decide whether a call
is repeated.

The flag stays off. Cancellation is still not forwarded across the hop and an
owner still never exits on idle, which are the two remaining gaps in #606.

Each rule is pinned by the mutation that breaks it and nothing else: dropping the
shield fails only the cancelled-caller test, dropping the generation check only
the late-failure test, dropping the shared task the concurrent one, keeping the
cache the annotation test, and clearing it on adoption the held-listing test. A
client driven against a real proxy server covers what none of those can: with the
middleware left unregistered, only those tests fail. The full suite, ruff and ty
pass.

See also: #606

## Synthetic prompt

> A proxy is pinned to the owner it attached to at startup, so an upgrade that
> stands the old owner down strands it. Make it recover. Classify a transport
> failure where the boundary is known, tagging it with the owner that failed and
> whether the request had left the process, then elect a replacement on failure
> and repeat the call when repeating it is safe: always for a listing, for a
> mutating tool only when nothing was sent, and otherwise only when the tool
> declares `readOnlyHint`. Reuse `daemon_auth`'s annotation rule rather than
> writing a second one. Run the election in a thread, one at a time, shielded
> from a caller that gives up. Make sure a stale component cache cannot decide
> whether a call may be repeated.

### DIFF
--- a/linkedin_mcp_server/daemon_auth.py
+++ b/linkedin_mcp_server/daemon_auth.py
@@ -306,7 +306,7 @@ class FrontendAuthRepairMiddleware(Middleware):
             logger.info("Signed in; not repeating a call that had already started")
             return result
 
-        if await _the_tool_changes_something(context):
+        if await a_repeat_could_change_something(context):
             # Repaired, but not run again. The auth is fixed and the next call
             # will work; what this one gets is the owner's own message, which
             # already says to retry.
@@ -350,10 +350,15 @@ class FrontendAuthRepairMiddleware(Middleware):
             return result
 
 
-async def _the_tool_changes_something(
+async def a_repeat_could_change_something(
     context: MiddlewareContext[mt.CallToolRequestParams],
 ) -> bool:
     """Whether running this tool again could repeat an effect on LinkedIn.
+
+    Public because two middlewares need the same answer: this module replays a
+    call after signing in, and ``daemon_proxy`` replays one after electing a
+    replacement owner. One rule, in one place, so the two cannot drift apart in
+    the unsafe direction.
 
     Read from the tool's own annotations, which survive the hop
     (``ProxyProvider`` copies them), rather than from a list of names kept here:
--- a/linkedin_mcp_server/daemon_proxy.py
+++ b/linkedin_mcp_server/daemon_proxy.py
@@ -12,37 +12,41 @@ them is a way to leak a credential or to hang a call.
 No client *session* is cached: one is built per upstream operation, because that
 is how ``ProxyProvider`` uses its factory and because a single shared session
 would outlive the owner it was opened against. The address and token are read
-per operation too, from :class:`DaemonProxyBackend`, so following a replacement
-takes no further machinery — see below for what it still takes.
+per operation too, from :class:`DaemonProxyBackend`, so a proxy follows a
+replacement owner instead of keeping an address it captured at startup.
 
-**A proxy is still pinned to the owner it started with, and an upgrade is enough
-to strand it.** Not because the address is captured any more, but because
-nothing replaces the attachment the backend holds. A proxy whose owner goes away
-therefore fails every operation afterwards rather than finding the replacement.
-Reproduced end to end: a proxy serving 19 tools was asked nothing, its owner was
-told to stand down the way a newer build does
+**Which owner that is changes while this process runs**, and the rest of this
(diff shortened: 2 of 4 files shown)


stickerdaniel:
* **#649** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/649?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/649?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#648** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/648?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

stickerdaniel:
### Merge activity

* **Aug 1, 1:31 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/649).
* **Aug 1, 1:32 PM UTC**: [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/649) rebased this pull request as part of a merge.
* **Aug 1, 1:34 PM UTC**: @stickerdaniel merged this pull request with [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/649).

PR #651, opened 2026-08-01
feat(daemon): Stop a call once nobody is waiting for it

stickerdaniel:
Cancellation does not cross the daemon's loopback hop, so a frontend that gives
up leaves the owner driving the browser for nobody. Measured in this repository:
the client was answered at 0.66s with an error and the effect landed 0.7s later,
against a LinkedIn account nobody was watching. The owner cannot tell the
difference, because a call somebody abandoned looks exactly like one somebody
wants.

The frontend now says it is still waiting, every two seconds, on an
authenticated control route, and the owner cancels a call it has not heard about
in ten. The identifier travels in a header rather than in MCP call metadata,
because it has to reach the owner on the heartbeat as well as on the call and
only one of those is an MCP message. A call and its heartbeats are bound to one
owner together: with no component cache every call re-lists first, so a
replacement adopted in between would otherwise send the request to one owner and
the beats to another, and the owner running the call would cancel it for silence
it never had a chance to hear.

The middleware sits outside the serializing one. A call spends most of its life
queued, and a frontend that gives up while its call is still waiting its turn is
the commonest way to end up with work nobody wants; inside, a call would only
become cancellable once it already had the browser.

A stall in the owner is not abandonment. A gap between the owner's own expiry
scans of ten times its poll interval means the process itself was not
scheduled — a slept laptop, heavy load, a long synchronous stretch — and every
clock restarts rather than expiring calls whose frontends were beating
throughout. A shutdown that races the expiry is reported as a shutdown, told
apart by the same `cancelling()` counter `browser_lifespan` already uses.

There is no `PROTOCOL_VERSION` bump, and both invariant comments now say why
instead of being quietly contradicted. A bump makes an old owner unreadable, and
an owner nobody can read is one nobody can ask to stand down, so the turnover an
upgrade depends on would break for every owner already installed. The additions
are optional in both directions instead: a new frontend meeting an old owner
gets 404 and proceeds unheard, and an old frontend sends no marker, which counts
as active and is never cancelled. Both directions are tested rather than
asserted.

The cadence and the expiry were measured rather than chosen. A heartbeat round
trip costs 0.9ms against an idle owner and 1.8ms while it drives Chromium; a
detached process was at most 3.6ms late over 25 beats; and the longest the
owner's synchronous extraction holds its event loop is 130ms, on a page far
larger than a real one. Ten seconds is five cadences and about seventy times the
worst delay measured, and the test pins the relationship rather than the numbers.

Each rule is caught by one mutation and nothing else: dropping the route's token
check, dispatching before the first beat, leaking the beater task, one
process-wide identifier, watching an unmarked call, following the backend to a
replacement mid-call, treating 404 as support, stripping the header in the
client factory, and omitting the expiry from the owner's serving loop. Two of
those tests were vacuous when first written, and only the mutations said so. The
full suite, ruff and ty pass.

See also: #606

## Synthetic prompt

> A frontend that gives up leaves the daemon owner scraping, because
> cancellation is not forwarded across the hop. Add heartbeats: an authenticated
> `POST /control/heartbeat` on the owner, a call identifier in a header that
> reaches the owner on both the beat and the call, a frontend middleware that
> beats while it waits and stops the moment it does not, and an owner middleware
> outside the serializing one that cancels a call nobody has asked about.
> Measure the cadence and the expiry instead of choosing them. Decide the
> `PROTOCOL_VERSION` question explicitly and make the invariant comments say
> what you decided.

### DIFF
--- a/linkedin_mcp_server/daemon_descriptor.py
+++ b/linkedin_mcp_server/daemon_descriptor.py
@@ -69,12 +69,24 @@ from linkedin_mcp_server.private_state import harden_directory, harden_file
 #: recognise the value refuses rather than guessing at fields it cannot read.
 SCHEMA_VERSION = 2
 
-#: Bumped when the daemon's own protocol changes: the control routes, the call
-#: metadata, or the ping contract. Compatibility keys on this rather than on the
-#: package version, because the documented install is ``@latest`` and this
-#: project releases often. Matching exact versions would mean a manual shutdown
-#: after almost every release, which is precisely the friction the daemon exists
-#: to remove.
+#: Bumped when the daemon's own protocol changes incompatibly: the control
+#: routes, the call metadata, or the ping contract. Compatibility keys on this
+#: rather than on the package version, because the documented install is
+#: ``@latest`` and this project releases often. Matching exact versions would
+#: mean a manual shutdown after almost every release, which is precisely the
+#: friction the daemon exists to remove.
+#:
+#: **Incompatibly is doing the work in that sentence, and it was added rather
+#: than assumed.** The heartbeat route and the call marker
+#: (:mod:`linkedin_mcp_server.daemon_liveness`) are a control route and call
+#: metadata both, and they did not bump this. A bump makes an old owner
+#: unreadable — :func:`read` refuses the file before an attachment exists — and
+#: an owner nobody can read is an owner nobody can ask to stand down, so the
+#: turnover that an upgrade depends on would break for every owner already
+#: installed. An addition that both sides can do without is therefore not a
+#: reason to bump; one that either side would misread is. The test for it is
+#: whether the two mixed pairings still work, and for the heartbeat they are
+#: tested rather than asserted.
 PROTOCOL_VERSION = 1
 
 _DESCRIPTOR_FILE = "daemon.json"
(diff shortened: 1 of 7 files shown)


stickerdaniel:
* **#652** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/652?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#651** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/651?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/651?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#650** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/650?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

stickerdaniel:
### Merge activity

* **Aug 1, 5:46 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/651).
* **Aug 1, 5:51 PM UTC**: [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/651) rebased this pull request as part of a merge.
* **Aug 1, 5:52 PM UTC**: @stickerdaniel merged this pull request with [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/651).

PR #652, opened 2026-08-01
feat(daemon): Let an owner exit once nothing needs it

stickerdaniel:
An owner closes the browser when it goes idle and then keeps the daemon lock for
the machine's uptime. What the next election waits for is the process, not the
Chromium it stopped running hours ago.

It now exits on a third condition, after wedge and turnover, reusing
`browser_idle_timeout_seconds` rather than adding a second dial: that setting
already says how long the owner keeps the browser, and a separate one would let
the browser close while the process stayed, which is the same gap with an extra
knob on it. Zero still disables it.

The idle clock starts where the descriptor is published, not at startup. Before
that nobody can reach this owner, and counting the import and the browser launch
as quiet time would let a short timeout fire before the frontend that asked for
the owner could call it.

Every call holds the door open, marked or not. An unmarked call cannot be
cancelled, but calling it idleness would exit in the middle of one and cut off
exactly the older frontends the heartbeat change promises to keep serving. That
accounting is new here; the previous change tracked only the calls it could
identify.

The stale descriptor is left behind deliberately. The next election probes it,
is refused, and elects a replacement, whereas deleting it here would race
whoever publishes next.

The real-process test is the one that matters, and it earned its place twice.
Every in-process rule stayed green with the clock never started at the publish
site, because each of them drives the tracker directly. And the first version of
that test watched an owner running on the 600 second default, because the
election suite's harness builds an `AppConfig()` and ignores the environment:
the feature worked and the test was lying. It elects through its own script now
and fails if either the publish-site call or the exit branch is removed.

One redundancy is documented rather than removed: the in-flight check inside
`quiet_for()` is belt-and-braces, since `call_started()` already clears the
clock, and removing it breaks no test. It should not be read as a rule anything
verifies.

The full suite, the slow test, ruff and ty pass.

See also: #606

## Synthetic prompt

> The daemon owner never exits on idle: it closes the browser and keeps the lock
> for the machine's uptime. Add a third exit condition to its serving loop after
> wedge and turnover, reusing `browser_idle_timeout_seconds` with zero as the
> off switch. Start the idle clock where the descriptor is published rather than
> at startup, and count every call as activity including the ones with no
> heartbeat marker, since those cannot be cancelled but are not idleness either.
> Leave the stale descriptor for the next election to probe. Cover it with a
> real detached process, because nothing in one interpreter can see the wiring.

### DIFF
--- a/linkedin_mcp_server/daemon_liveness.py
+++ b/linkedin_mcp_server/daemon_liveness.py
@@ -134,6 +134,18 @@ class CallLiveness:
 
     _waiting: dict[str, _Waiting] = field(default_factory=dict)
 
+    #: How many calls are running, marked or not. Separate from ``_waiting`` on
+    #: purpose: an unmarked call cannot be cancelled, but it is emphatically not
+    #: idleness, and an owner that exited during one would cut off the very
+    #: frontends this build promises to keep serving.
+    _in_flight: int = 0
+
+    #: When this owner last did anything for anybody, or ``None`` while it is
+    #: still starting. Idleness is only meaningful once the endpoint is
+    #: published: before that nobody could have called, and an owner that
+    #: counted its own startup as idle time would exit during it.
+    _quiet_since: float | None = None
+
     #: When the owner last got round to asking who was still waiting. Compared
     #: against the next scan so a stall in this process is not charged to the
     #: frontends.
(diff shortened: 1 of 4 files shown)


stickerdaniel:
* **#652** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/652?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/652?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#651** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/651?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#650** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/650?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

stickerdaniel:
### Merge activity

* **Aug 1, 5:46 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/652).
* **Aug 1, 5:53 PM UTC**: [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/652) rebased this pull request as part of a merge.
* **Aug 1, 5:55 PM UTC**: [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/652) couldn't merge this PR because it was not satisfying all requirements (Failed CI: 'test').

PR #654, opened 2026-08-01
ci(release): Split build and publish jobs

### LINKED ISSUE #653: [FEATURE] Split release workflow so build code never shares a job with publish credentials
### Feature Description

The auto-release workflow runs `uv build`, `bunx @anthropic-ai/mcpb`, and `docker build` inside the same job that holds `id-token: write` for PyPI Trusted Publishing, plus `GH_ADMIN_TOKEN` (which removes branch protection), Docker Hub credentials, and `contents: write`.

Any code that executes during that job can mint an OIDC token for the project and exchange it at PyPI for an upload token. The build backend is `setuptools>=68.0` plus `wheel`, unpinned and resolved from PyPI at build time (`uv.lock` does not cover `[build-system].requires`), and the MCPB CLI is invoked through `bunx` without a version. A compromised release of any of those would run in a job that can publish as this project or tamper with `dist/` before upload.

### Use Case

Supply-chain hardening. Flagged by @webknjaz (maintainer of `pypa/gh-action-pypi-publish`) in https://github.com/stickerdaniel/linkedin-mcp-server/pull/512#discussion_r3691116168. The Python packaging guide has always prescribed splitting build and publish for exactly this reason.

### Suggested Approach

Split the release job so no job both executes third-party code and holds a credential it does not need:

- `build` produces the distributions with no credentials and uploads them as an artifact
- `publish-pypi` downloads that artifact and publishes, with `id-token: write` as its only permission, no checkout, bound to a `pypi` environment
- `prepare-release` writes the version, commits, and tags, running only git/gh/sed
- `docker` and `build-mcpb` build from the released commit, holding only the credentials their own registry push needs
- `create-github-release` publishes the release from downloaded artifacts

Also pin the build backend for CI through a hashed `uv build --build-constraint`, keep `[build-system].requires` a lower bound so published sdists stay buildable, drop the obsolete `wheel` requirement, and pin the MCPB CLI version.

stickerdaniel:
Closes #653

The release job built the distributions, packed the MCP bundle and built the image while holding the PyPI OIDC identity, the admin token and the Docker Hub credentials. Any code running there, including a build backend resolved fresh from PyPI on every release, could mint a publishing token for this project or tamper with the artifacts before they were uploaded. @webknjaz flagged this on #512.

Each of those steps now runs in its own job, so no job holds a credential its own work does not need. The distributions are built without any publishing credentials and reach the publishing job as an artifact; that job has `id-token: write` as its only permission, checks nothing out, and runs nothing but the upload. Writing the version back, tagging, pushing the image and packing the bundle follow from there, each with only what it needs, and the Docker and bundle jobs build from the commit that carries the updated manifest. The workflow now also declares `contents: read` at the top instead of silently inheriting write, and drops a `packages: write` scope nothing was using.

Alongside that, CI builds pin the build backend through a hashed constraint file, while the requirement in `pyproject.toml` stays a lower bound so published sdists remain buildable by anyone. The `wheel` requirement is gone since setuptools has bundled `bdist_wheel` since 70.1, and the MCPB CLI is pinned like every other dependency, both kept current by Renovate. The image build is not covered by that pin, since `uv sync` has no way to take one; #655 tracks it.

Verified locally: the constrained build resolves and hash-checks setuptools 83.0.0 and produces both distributions, `uv lock --check` is unchanged, zizmor reports nothing above informational, and a per-job audit confirms no job holds a credential it does not use.

## Synthetic prompt

> The release workflow builds the package in the same job that holds `id-token: write` for PyPI Trusted Publishing, plus an admin token and Docker credentials. Split it per the PyPA guide so no job both runs third-party code and holds credentials it doesn't need, pin the build backend with hashes for CI without constraining published sdists, and pin the MCPB CLI.


stickerdaniel:
* **#654** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/654?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/654?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
@greptileai review

PR #656, opened 2026-08-01
test(daemon): Ask the child, not the worker, who holds the lock

stickerdaniel:
`test_an_adopted_lock_does_not_leak_to_later_children` failed intermittently in
the parallel job on Linux: three times in about eight runs across two branches,
always passing when re-run, and never failing when the file was run on its own.
It was not reproducible in seven local configurations, including Linux at the
runner's worker count under the exact CI command.

**The mechanism is not established, and this does not claim one.** The lock path
is isolated twice over — the state root lives under the test's own `tmp_path`,
and the directory under it is keyed by a hash of the auth root's device and
inode — so two tests cannot collide on it, and an ordinary open descriptor does
not block `flock` either.

What can be said is where the test asked its question. `flock` is held against
the inode, so "can anyone take this lock" is answered by every holder anywhere,
not only by the descriptor under test. Asked from a pytest worker that has run
thousands of other tests, in a session whose composition decides what that
worker has open, the answer is about the worker as much as about the adoption.

So the scenario now runs in a clean interpreter: take the lock, hand a copy on,
adopt it, launch an unrelated child that inherits descriptors, let go, and ask
whether anyone can take the lock. Nothing but this scenario exists in that
process. The suite already had a helper for exactly this, used by the ownership
tests.

Alongside it, the invariant is asserted directly: after `adopt()`, the
descriptor is no longer inheritable. That one spawns nothing and contends for
nothing, so there is nothing about it left to disturb.

Two mutations, two failures. Removing `set_inheritable(False)` from `adopt()`
fails both. A `try_acquire()` that refuses whenever the lock file already exists
fails the end-to-end one, which is why that half asks through the production API
rather than a raw `flock`: an earlier draft used `flock` and accepted that
implementation.

Verified stable across ten local runs of the pair, the full local suite, and
five Linux runs at the runner's worker count.

See also: #606

## Synthetic prompt

> A test in `tests/test_daemon_lock.py` flakes on Linux CI under xdist and
> cannot be reproduced locally. Work out why it is fragile rather than retrying
> it, and replace it with tests that cover the same property deterministically.
> Do not assert a mechanism you could not reproduce. Prove the replacement still
> catches the bug it was written for by removing the fix and watching it fail.

### DIFF
--- a/tests/test_daemon_lock.py
+++ b/tests/test_daemon_lock.py
@@ -512,39 +512,93 @@ class TestFork:
             os.waitpid(pid, 0)
 
 
+#: The whole adoption, in one clean interpreter: take the lock, hand a copy on,
+#: adopt it, launch an unrelated child that inherits descriptors, let go, and ask
+#: whether anyone can take the lock now.
+#:
+#: All of it in a child rather than in the test, and that is the point. `flock`
+#: is held against the inode, so *any* holder anywhere answers this question, not
+#: only the descriptor under test. Asked from a pytest worker that has run
+#: thousands of other tests, the answer is about that worker as much as about the
+#: adoption. A fresh interpreter holds only what this scenario gave it, which
+#: takes the worker out of the result — whether or not the worker is what made
+#: the earlier version of this test fail intermittently on Linux. That was never
+#: reproduced and is not claimed here.
+_ADOPT_THEN_LAUNCH_AN_UNRELATED_CHILD = """
+    import os
+    import subprocess
+    import sys
+    from pathlib import Path
+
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#656** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/656?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/656?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #666, opened 2026-08-03
fix(browser): Remove the user-agent override

stickerdaniel:
The import synthesized the source browser's user agent and every later session
replayed it, on the reasoning that LinkedIn binds a session token to the
fingerprint it was minted under. That was never measured and no source says it.
What the mechanism measurably did instead: Patchright overrides only the UA
string, so the client hints kept reporting the real browser and any page could
ask the same question twice and get two answers. Measured here as Chrome 143 in
the user agent against Chromium 148 in the hints. The override never reached
service workers at all ([playwright#5237](https://github.com/microsoft/playwright/issues/5237),
closed as an upstream Chromium bug), the value was frozen at import time while
the browser it described kept updating, and on Apple Silicon it flipped
`sec-ch-ua-arch` to `x86` while WebGL still named an Apple GPU.

A browser telling the truth about itself beats one caught lying, so the
synthesis, the `chromium_versioned` and `ua_brand_suffix` registry inputs that
fed it, the `user_agent` key in `source-state.json`, and both settings are gone.
`USER_AGENT` and `--user-agent` are refused at startup rather than ignored:
someone who set them wanted the browser to present itself differently, and
starting anyway would leave them believing it still applies. `BrowserManager`
refuses a `user_agent` in its launch options for the same reason, since that
dict is spread straight into the context options and a stray key would take
effect with nothing in between noticing.

`BrowserConfig.user_agent` stays as a field nothing reads. `SHARED_CONFIG_FIELDS`
hashes it into the daemon's configuration fingerprint, and dropping a field
there makes a running owner unreadable to a client of the other version rather
than merely mismatched. It goes once owner turnover survives a fingerprint
change.

Existing sessions will present a different user agent after upgrading, and
nothing is built for that: a browser going from 143 to 150 is ordinary, while a
user agent contradicting its own client hints is not. State files written before
this still load, because loading filters to the fields the dataclass declares,
and they shed the stale key on the next login.

Full suite green at 1648 passed; the three new guards were each verified red
against a mutation that reintroduces the override.

See also: #606

## Synthetic prompt

> Remove the synthesized user-agent override entirely. Delete
> `browser_import/user_agent.py` and the `chromium_versioned` / `ua_brand_suffix`
> registry fields that only fed it, drop `SourceState.user_agent` and the
> `write_source_state` argument, and strip the `user_agent` plumbing from
> `BrowserManager`, `drivers/browser.py` and the manual login. Refuse
> `USER_AGENT` and `--user-agent` at startup with an error explaining that the
> override changed the UA string but not the client hints and never reached
> service workers. Keep `BrowserConfig.user_agent` as an unread field so the
> daemon configuration fingerprint is unchanged, and say why in the code. Add
> tests that fail if any of it comes back, and update the docs.


stickerdaniel:
* **#666** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/666?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/666?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #667, opened 2026-08-03
docs(docker): Publish HTTP on loopback

stickerdaniel:
The documented Docker HTTP command published `-p 8080:8080`, which puts an MCP endpoint with no authentication on every interface of the host. The mistake is silent: it works perfectly for whoever pasted it, and nothing tells them the LinkedIn session behind it is now reachable from the network.

The server cannot catch this on its own. A container has to bind the wildcard, because a process bound to `127.0.0.1` inside one is unreachable through a published port at all, so exposure is decided entirely by the publish address on the host. From inside, `-p 127.0.0.1:8080:8080` and `-p 8080:8080` look identical. Container detection is no way out either, since `LINKEDIN_MCP_CONTAINER` is a full override and therefore not a security boundary. Refusing a non-loopback bind, which was the first idea, would have broken the documented command while protecting nobody.

So the documentation carries it. The example publishes to loopback and explains why both halves are needed, and the warning now names the non-Docker remedy, the Docker remedy and the SSH tunnel instead of only suggesting a host value that is wrong inside a container. It also states plainly that host and origin checking is not access control: it stops a website from pointing a name at this server, not somebody who can already reach the address.

The test pins the published address, anchored on `docker run` rather than on the heading, because the README has three blocks headed "HTTP Mode Example" and only the containerised one is at risk. A heading-anchored test would have matched the uvx example and stayed green while the Docker command was wrong.

Real inbound authentication is the actual fix and is deliberately not in this change. Three comparable MCP servers bind `0.0.0.0` in Docker and gate on a bearer token; one refuses Streamable HTTP without one and permits unauthenticated only on real loopback. That is the shape to copy when it is built.

Full suite green at 1662 passed. Both mutations verified red: reverting the publish to `-p 8080:8080`, and "fixing" the host to `127.0.0.1` inside the container.

See also: #606

## Synthetic prompt

> The documented Docker HTTP command in README.md publishes on every interface,
> exposing the unauthenticated MCP endpoint. Change it to publish on loopback and
> explain why `--host 0.0.0.0` is still required inside the container. Sharpen the
> non-loopback bind warning in `config/schema.py` so it names the Docker remedy
> too, and says host/origin checking is not access control. Mirror it in
> `docs/docker-hub.md`. Add a test that pins the published address, anchored on
> the command rather than the heading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,8 +385,10 @@ Drop that prefix and Docker publishes on every interface, which puts an
 endpoint with no authentication on your network. The server cannot tell the two
 apart, so it warns either way.
 
-Note that other containers on the same host still reach it through
-`host.docker.internal`.
+Loopback publishing limits this to the machine, not to the container. Other
+containers on the same host can still reach it through `host.docker.internal`
+wherever that name resolves, which is the default on Docker Desktop and
+OrbStack but not on native Linux Docker.
 
 Runtime server logs are emitted by FastMCP/Uvicorn.
 
--- a/README.md
+++ b/README.md
@@ -372,10 +372,21 @@ This opens a browser window where you log in manually (5 minute timeout for 2FA,
 ```bash
 docker run -it --rm \
   -v ~/.linkedin-mcp:/home/pwuser/.linkedin-mcp \
-  -p 8080:8080 \
+  -p 127.0.0.1:8080:8080 \
   stickerdaniel/linkedin-mcp-server:latest \
   --transport streamable-http --host 0.0.0.0 --port 8080 --path /mcp
 ```
+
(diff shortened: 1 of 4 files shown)


stickerdaniel:
* **#667** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/667?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/667?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #669, opened 2026-08-04
fix(browser): Always launch full Chromium

stickerdaniel:
Without a `channel`, Playwright picks the browser *binary* from the `headless` flag alone. In the installed patchright's `coreBundle.js`, `getExecutableName()`:

```js
return options2.headless ? "chromium-headless-shell" : "chromium";
```

`--login` forces `headless=False` and so minted every session in the full Chrome for Testing, while scraping defaults to `headless=True` and used the stripped headless shell, which has no plugins, no `window.chrome` and an incoherent notification permission. **Every user's session was created in one browser and used in another.**

Naming `channel="chromium"` resolves the same registry entry in either mode. Measured after the change, over CDP `Browser.getVersion`:

```
headless=True   product Chrome/148.0.7778.96   UA ...HeadlessChrome/148.0.0.0...
headless=False  product Chrome/148.0.7778.96   UA ...Chrome/148.0.0.0...
```

Same binary now. The second column is the part this PR deliberately does **not** fix: the headless token comes from the mode, not the executable, so no change to which binary runs can remove it. That is separate work.

`CHROME_PATH` still wins outright and is passed alone. Patchright would let the path win over a channel anyway, but leaving two competing selectors in one options dict is a trap for whoever reads it next.

Bootstrap had to move with the launch or the two would disagree into an unbounded loop: a shell-only install read as ready, the launch demanded the missing full browser, `dependencies.py` invalidated only the metadata, and the next setup installed the shell again. Readiness now asks one question, whether the browser this server launches is installed, instead of branching on a flag that selects a mode. `full_chromium_ready` required *every* install target and so failed forever on a full-only install; it is now `browser_ready` and checks the browser.

Setup installs in one stage. The download effect cuts both ways, measured against the CDN for 148.0.7778.96 mac-arm64: 170 MiB, against 263 MiB for anyone who previously ended up needing the full browser, and against 92 MiB for the default headless user who only ever fetched the shell. The second comparison is the one users will notice, and the release notes will say so.

`eager_full_chromium` becomes inert: with one browser, "up front" and "lazily" describe the same install. The field stays in configuration and in `SHARED_CONFIG_FIELDS`, because `daemon.py` rejects a configuration-fingerprint mismatch *before* it compares package versions, so removing a field there makes a running owner of the other version unreadable and therefore impossible to retire.

Full suite green at 1658 passed. New builder tests verified red against both mutations: dropping the channel, and letting the managed choice override an operator's `CHROME_PATH`.

See also: #606

## Synthetic prompt

> Playwright picks the Chromium binary from the `headless` flag when no channel
> is set, so `--login` uses full Chrome for Testing while scraping uses the
> headless shell. Set `channel="chromium"` in the shared launch-options builder
> so both paths use one binary, keeping `CHROME_PATH` as an operator override.
> Then fix bootstrap to match, or a shell-only install loops: make readiness
> check the full browser only, stop branching on `headless`, install in one
> `--no-shell` stage, and drop the shell-only paths. Retire
> `eager_full_chromium` as behaviour but keep the field for the daemon
> fingerprint. Update the tests that encode the old two-stage install.


stickerdaniel:
* **#669** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/669?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/669?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #671, opened 2026-08-05
fix(bootstrap): Skip the install for a custom Chrome

stickerdaniel:
`ensure_browser_installed()` downloads the managed browser for `--login`, `--status` and `--import-from-browser` even when `CHROME_PATH` points at the executable those modes will actually launch. `_uses_custom_chrome()` already existed and its docstring already said why the download is pointless; this caller simply never asked.

It mattered less before #669, when two of the three modes needed only the 92 MiB headless shell. They all want the full browser now, so it is 170 MiB fetched for something that is never run. For an operator whose network cannot reach the CDN it is worse than waste: with a perfectly good browser on disk they can neither sign in nor check status.

Also corrects two comments that outlived the change they described. The headless shell is no longer "the only binary the default path launches", and there are no longer two install stages to serialise.

Found by review of #669 rather than by a test, which is itself the gap: the custom-Chrome path was covered for the background gate and for `_run_login_flow`, but not for this CLI wrapper. It is now, verified red against removing the short circuit.

One thing worth flagging for the next reader. Adding the check means this function reads the configuration, and `get_config()` parses `sys.argv` lazily, which pytest cannot resolve on its own. Three existing tests started failing for exactly that reason, and they now state which world they are in rather than inheriting one. The same trap is documented on `build_launch_options`.

Full suite green at 1660 passed.

See also: #606

## Synthetic prompt

> `ensure_browser_installed()` downloads the managed Chromium for the CLI modes
> even when `CHROME_PATH` is set, so an operator with their own browser and no
> CDN access cannot log in. Short-circuit it with the existing
> `_uses_custom_chrome()`, add tests for both directions, and fix the comments
> that still describe the headless shell as the default binary and the install
> as two stages.

### DIFF
--- a/linkedin_mcp_server/bootstrap.py
+++ b/linkedin_mcp_server/bootstrap.py
@@ -65,11 +65,12 @@ _REGISTRY_NAME_TO_DIR_PREFIX = {
     "chromium-headless-shell": "chromium_headless_shell-",
 }
 
-# On-disk dir prefix of the headless shell — the only binary the default
-# headless scrape + auto-import path launches.
+# On-disk dir prefix of the headless shell. Nothing launches it any more —
+# every launch names ``channel="chromium"`` — but the prefix is still needed to
+# recognise one in an install written before that change.
 _SHELL_DIR_PREFIX = "chromium_headless_shell-"
-# On-disk dir prefix of full Chrome for Testing — needed only for the headed
-# interactive-login fallback or an operator-configured --no-headless run.
+# On-disk dir prefix of full Chrome for Testing: the browser this server runs,
+# in either mode.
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#671** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/671?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/671?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #672, opened 2026-08-05
feat(browser): Work in a window that never opens

stickerdaniel:
The browser announced itself as headless, and the announcement comes from the *mode* rather than the binary: Chromium prepends the bare string `Headless` to the product name at runtime, so both the user agent and the `sec-ch-ua` brands read `HeadlessChrome`. Naming a channel cannot remove that, which is why #669 deliberately did not claim to. But this server must not put a window on somebody's desktop either.

So where the platform allows it, the browser launches headed and works in a target Chromium creates with no UI window. `BrowserManager.headless` keeps its public meaning, "no visible window"; it is simply no longer how that is achieved.

Measured end to end through the production path on macOS:

```
page url  : about:blank#63dbff0c723048d6
UA        : ...Chrome/148.0.0.0...        no HeadlessChrome
brands    : Not/A)Brand, Chromium         webdriver: false
outer 1280x720  screen 1280x720  fits     visible / focused
rAF       : 122/s   control window: 122/s
cookie across a full restart: survives
```

The rAF figure is the one that mattered. Hiding the application at OS level, the obvious alternative, throttled it to about 1 Hz against 120 and un-hid itself on the next `new_page()`. A hidden target runs at the same rate as an ordinary window.

## It applies to macOS only, and that is measured rather than scoped

Two things have to be true, and only one is about displays.

A headed browser needs a display server. Measured in the published container image: with no `DISPLAY`, `launch_persistent_context(headless=False)` fails with a `TargetClosedError` before any of this code runs.

And the browser has to survive losing its last visible window, since removing that window is the whole point. Measured in the same image under Xvfb — the hidden target *is* created and found, and then:

```
close startup page : FAIL   Chromium exits, the hidden page dies with it
keep startup page  : OK
```

Linux quits a browser with its last visible window. macOS does not, which is why it works there. Windows is untested and plausibly behaves like Linux, so it is not claimed. Everything else falls back to Chromium's headless mode and logs why.

Linux is less a gap than a different answer: under a virtual display nobody is looking at the screen, so an ordinary window is already invisible and there is nothing to hide. That is the container's path and it belongs with the change that gives the container a display.

## Three details decide whether it works at all

Each was measured, and a wrong one does not raise — it hands back the wrong page, or a page that vanishes later.

**Created from a browser-level CDP session.** Through `context.new_cdp_session(page)`, the obvious call, the target dies when that page closes: the browser stays up, `new_page()` still works, and the hidden page is gone.

**Found by a nonce, never by position.** The attach flag promotes every `other` target, and a real run had a component extension's page in `context.pages` beside the one we asked for.

**`about:blank#<nonce>`, not a loopback origin.** With a proxy configured and no bypass, a loopback URL is fetched *through* the proxy and the page lands on `chrome-error://chromewebdata/`. `about:blank` has no origin to route.

No bootstrap extension is needed: Chromium's own component extension already supplies the non-visible frame target its precondition wants. That removes `--load-extension`, so `CHROME_PATH` pointing at branded Chrome keeps working — measured, since Chrome 137+ ignores that switch.

The attach flag is set only for launches that create a hidden target. It lives in the driver process for that process's whole life, so restoring the parent environment does nothing to the child; a visible login would otherwise spend its entire run promoting `other` targets. It also survives overlapping launches, which it previously did not: process-global state mutated across an `await` left it set in two runs out of five.

## What this does not fix

A window is on screen for about half a second on every browser start, median 553 ms over five runs, and it cannot be shortened from here: roughly 250 ms passes before `launch_persistent_context()` returns, about 340 ms is macOS tearing the window down, and about 90 ms is ours. Zero would need `--no-startup-window`, which hangs Playwright for its full timeout ([playwright#42093](https://github.com/microsoft/playwright/issues/42093), closed as unsupported args).

The windowless page also reports `outerWidth == screen.width`, which no real window does. Unchanged from the previous headless default rather than introduced here, and recorded in `docs/browser-fingerprint.md` as still worth fixing.

Full suite green at 1686 passed. Every new assertion was verified red against the mutation it exists to catch, including three that originally were not: the launch contract stayed green with `open_hidden_page()` deleted from `start()`, the helper's fake surfaced its page synchronously and so hid every ordering mistake, and the ordering assertion checked that `createTarget` had been *sent* before the close rather than that the page *existed*.

See also: #606

## Synthetic prompt

> Chromium's headless mode makes the browser announce itself via a `Headless`
> product prefix, so run headed and hide the page in a CDP target instead:
> `Target.createTarget({hidden: true})` on `about:blank#<nonce>` from a
> **browser-level** session, found by nonce rather than position, with
> `PW_CHROMIUM_ATTACH_TO_OTHER` scoped to launches that need it and safe against
> overlap. Measure which platforms can actually do this rather than assuming:
> a headed launch needs a display, and the browser must survive losing its last
> window. Fall back to real headless elsewhere and log why. Test the launch
> contract end to end, not just the helper.

### DIFF
--- a/docs/browser-fingerprint.md
+++ b/docs/browser-fingerprint.md
@@ -77,8 +77,35 @@ headless".
 | Bundled Chromium, headless, UA claiming 143 | 33% / 88% | `hasMissingChromeObject` high severity; UA and hints disagree |
 | Full Chromium headless (`channel="chromium"`) | 67% / 50% | Two high-severity fpscanner rules from the headless token |
 | Headless shell (the old default) | — | `plugins.length = 0`, no `window.chrome`, notification permission incoherent |
+| Hidden target, windowless mode | — | No headless token in UA or brands; `visible` / focused; rAF at 100% of a control window |
 | Docker, headed under Xvfb | 0% / 44% | No headless token, native hints; needs `xauth` |
 | Docker, Xvfb + Mesa llvmpipe | 0% / 44% | Restores WebGL; renderer string is a known software renderer |
+
+The windowless mode, measured end to end through `BrowserManager`:
+
+| | Value |
+|---|---|
+| User agent | `…Chrome/148.0.0.0…`, no `HeadlessChrome` |
+| `sec-ch-ua` brands | `Not/A)Brand`, `Chromium` |
+| `navigator.webdriver` | `false` |
+| `document.visibilityState` / `hasFocus()` | `visible` / `true` |
+| `requestAnimationFrame` | 122/s against a control visible window at 122/s |
+| Cookie across a full restart | survives |
+| Window on screen once settled | none |
+| Window on screen during startup | ~550 ms, median of five runs (504-593) |
+
+The rAF figure is the one that mattered: hiding the application at OS level
+throttled it to about 1 Hz against 120, which is what disqualified that
+approach. A hidden target runs at the same rate as an ordinary window.
+
+Two things this does not claim. The half second of visible window on every
+browser start cannot be shortened from here — roughly 250 ms passes before
+`launch_persistent_context()` returns and about 340 ms is macOS tearing the
+window down, leaving about 90 ms that is ours. And the windowless page reports
+`outerWidth == screen.width`, which no real window does, since a real one has
+chrome and sits inside its display. That is unchanged from the previous
+headless default rather than introduced here, and it stays on the list of
+things worth fixing.
 
 Window geometry, read from a page's own `<script>` on a loopback origin:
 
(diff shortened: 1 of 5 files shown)


stickerdaniel:
* **#672** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/672?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/672?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

PR #676, opened 2026-08-05
fix(auth): Refuse a profile root nobody claimed

stickerdaniel:
`USER_DATA_DIR` accepts any path, and six operations move or recursively delete what it names. Three of them reach the parent directory as well, because the portable cookies, the source metadata and the derived runtime profiles are siblings of the profile rather than children of it. A mistyped path therefore costs a directory nobody meant to name, and the emptiness of the profile proves nothing about what sits beside it.

Every destructive operation now goes through one guard, and it runs before any exists-or-empty short-circuit: a foreign directory holding none of our artifacts reached those early returns without ever being judged. The canonical default is ours by definition, which is what keeps the container working, since a path recorded on a host can never equal `/home/pwuser/.linkedin-mcp/profile`. Any other root needs a `profile-claim.json` marker in its auth root, written when the root is empty, when a `source-state.json` already living there names it, or when the operator says so once with `--claim-profile-root`. The marker sits outside the auth-state targets and outside the quarantine glob, because rotation and logout are the moments the next run needs it most, and its lock is its own rather than the browser lease: that lease is held for as long as a Chromium is open, and reaching for it here made starting the server depend on nobody else browsing.

This also pairs `expanduser` with `resolve` everywhere they had drifted apart. Split, a symlinked profile is moved lexically while its cookies are computed from the resolved parent, which leaves one session across two roots and reports nothing.

Upgrade impact is one group: anyone with a non-default `USER_DATA_DIR` whose parent already holds unrelated files and who has never signed in. They need `--claim-profile-root` once. The default carries itself, and an existing `source-state.json` migrates silently.

Verified by mutation: eighteen mutations, each removing one behaviour the tests name, and all eighteen fail the suite. Full suite 1743 passed, 10 skipped.

## Synthetic prompt

> `USER_DATA_DIR` accepts any path and several operations recursively delete it and its parent, so a typo destroys unrelated data. Add an ownership marker (`profile-claim.json`) written into the auth root, route every destructive operation in `session_state.py` through one guard that refuses a root we cannot prove we own, and always accept the canonical default so Docker keeps working. Give the operator a one-time `--claim-profile-root` escape. Fix the expanduser/resolve split while you are in there. Tests must be mutation-verified.

Generated with Claude Opus 5 high


stickerdaniel:
* **#676** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/676?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/676?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
@greptileai review

PR #690, opened 2026-08-08
feat(post): Add schedule_post tool

### LINKED ISSUE #689: [FEATURE] Add schedule_post tool (LinkedIn native Schedule-for-later)
## Feature Description

Add a `schedule_post` tool that schedules a feed post through LinkedIn's native Schedule-for-later flow in the share composer: open the composer, set the scheduled date and time, type the post text, and confirm.

## Use Case

Scheduling through LinkedIn's own scheduler means the post is stored by LinkedIn and publishes at the chosen moment whether or not the MCP server is running — no local queue, no daemon requirement. Combined with an assistant that drafts content, this closes the write loop: draft → schedule → LinkedIn publishes.

## Suggested Approach

- Tool: `schedule_post(text, schedule_date "YYYY-MM-DD", schedule_time "HH:MM" 24h, confirm_schedule)` with the same explicit confirmation gating as `send_message`; `confirm_schedule=false` runs the full flow as a dry run and discards the draft.
- Open the composer by URL (`/feed/?shareActive=true`) rather than by clicking the label-identified "Start a post" trigger.
- The composer renders in an open shadow root (`#interop-outlet`, the SDUI interop surface); Patchright locators pierce shadow roots, so plain CSS locators work while `page.evaluate` querySelector does not.
- Set the schedule **before** typing: the schedule round-trip rebuilds the editor and drops typed text (measured).
- Locale-independent identification: clock button via LinkedIn's `data-test-icon="clock-medium"` test hook; date/time inputs via their stable semantic ids (`#share-post__scheduled-date`, `#share-post__scheduled-time`); date/time strings formatted by copying the dialog's own prefill format (it arrives prefilled with "today" in the profile's locale).

manasnisar:
Closes #689

## What

Adds a `schedule_post` tool that schedules a feed post through LinkedIn's **native Schedule-for-later** flow, so the scheduled post lives on LinkedIn's side and publishes at the chosen moment whether or not this server is running.

- `schedule_post(text, schedule_date "YYYY-MM-DD", schedule_time "HH:MM" 24h, confirm_schedule)`
- Same explicit gating as `send_message` (`destructiveHint: True`): `confirm_schedule=false` runs the **entire** flow as a dry run — LinkedIn parses the date/time and its rendering is returned in `composer_text` — then discards the draft.
- Rejects malformed dates/times and past datetimes before touching the browser.

## How (what live probing established)

- **The composer lives in an open shadow root** (`#interop-outlet`, LinkedIn's SDUI interop surface). Patchright locators pierce shadow roots ([documented](https://github.com/Kaliiiiiiiiii-Vinyzu/patchright-python#closed-shadow-roots)); `page.evaluate` + `querySelectorAll` does not. The flow therefore uses plain CSS locators only, and a comment in `extractor.py` warns future edits away from JS in this flow.
- **Opened by URL**, `/feed/?shareActive=true`, not by clicking the "Start a post" trigger whose only stable identity is label text.
- **Schedule before typing.** The schedule dialog round-trip rebuilds the editor and silently drops typed text (measured); text typed after the round-trip survives.
- **Locale independence** per the scraping rules:
  - Clock button via LinkedIn's own test hook `svg[data-test-icon="clock-medium"]` (icon name, not a translation).
  - Date/time inputs via stable semantic ids `#share-post__scheduled-date` / `#share-post__scheduled-time` (not `ember\d+`).
  - Date and time strings are formatted **by copying the dialog's own prefill** — it arrives prefilled with "today"/a suggested slot in the profile's locale, which doubles as a worked example of order, separator, and 12h/24h convention. Residual ambiguity (prefill day == month, non-English 12h locales) is documented in the helpers' docstrings; the flow verifies the pickers' accepted values and reports `schedule_rejected` rather than scheduling at a wrong time.
  - Dialog primary action via the repo's existing last-button convention, scoped to the dialog holding the date input (hidden video-player dialogs on the feed would otherwise shift the count).
- **Dry-run cleanup without the discard prompt.** Escape does not close this composer, and the discard-prompt buttons have no locale-independent identity — but an *empty* composer dismisses cleanly, so cleanup empties the editor first and closes via `button[data-test-modal-close-btn]`.

## Verification (live, end-to-end)

- Dry run via direct extractor call and via `tools/call` on a running streamable-http server: LinkedIn echoed `Posting Fri, Aug 14 at 11:45 AM` for `2026-08-14 11:45`, draft discarded, no dialogs left behind.
- Real run (`confirm_schedule=true`): returned `scheduled: true`, and the post was independently confirmed in LinkedIn's own **Scheduled posts** list (screenshot taken during verification showed the entry "Posting Tue, Aug 11 at 10:00 AM").
- `ruff check`, `ruff format`, `ty check`, pre-commit all pass; 13 new tests (9 for the two locale-format helpers, 4 for the tool wrapper) and the affected test files pass (338 tests).

## Synthetic prompt

> Add a `schedule_post` tool that schedules a feed post via LinkedIn's native Schedule-for-later composer flow. Probe the live composer first (it renders in the `#interop-outlet` shadow root — use piercing locators, not page.evaluate; open it via `/feed/?shareActive=true`; set the schedule before typing because the round-trip drops typed text). Identify controls locale-independently (`data-test-icon="clock-medium"`, the `#share-post__scheduled-*` input ids, last-button-primary convention) and format date/time by copying the dialog's own prefill. Gate with `confirm_schedule` like `send_message`, with a full dry-run mode that discards by emptying the editor before dismissing. Add tool + extractor method, unit tests, README/manifest entries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ This MCP server is **free** and **open source**, supported by [**Unipile**](http
 | `get_job_details` | Get detailed information about a specific job posting | working |
 | `get_feed` | Get recent posts from the authenticated user's home feed | working |
 | `search_posts` | Search posts/content globally by keyword (the "Posts" tab) with an optional recency filter (past-24h/past-week/past-month) | working |
+| `schedule_post` | Schedule a feed post via LinkedIn's native Schedule-for-later flow (requires confirmation; publishes even when this server is offline) | working |
 | `close_session` | Close browser session and clean up resources | working |
 
 <br/>
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -221,41 +221,93 @@ _SCHEDULE_TIME_INPUT_SELECTOR = "input#share-post__scheduled-time"
 _COMPOSER_DISMISS_SELECTOR = "button[data-test-modal-close-btn]"
 
 
-def _format_schedule_date(year: int, month: int, day: int, current_value: str) -> str:
-    """Format a date the way the schedule dialog's own prefill formats one.
+def _resolve_date_order(
+    current_value: str, placeholder: str, today: datetime.date
+) -> tuple[str, str, str] | None:
+    """Determine the date input's slot order, e.g. ``("m", "d", "y")``.
 
-    The date input arrives prefilled with *today* in the profile's locale
-    (measured: ``8/8/2026`` for en with placeholder ``mm/dd/yyyy``). Today's
-    components are known, so the prefill doubles as a worked example of the
-    expected order and separator — a measurement, not a translation table.
-    Only when today's day and month are equal (~1 day a month) is the order
-    unreadable from the example; then the en ``month/day/year`` order is
-    assumed. Locales whose *order* differs from en and whose prefill is
-    ambiguous that day may mis-order — a bounded, dated limitation.
+    Two independent measurements, tried in turn; None when neither decides.
+
+    1. The prefill is LinkedIn's rendering of its own "today", which can sit
+       at most one calendar day from the server's UTC today. Reading the two
+       small numbers both ways yields two *different* dates whenever day and
+       month differ, and two distinct readings are never both inside a
+       three-consecutive-day window — so a reading that lands in the window is
+       proof of the order, not a guess.
+    2. The placeholder's short tokens. Across the Latin-script locales
+       LinkedIn ships, the month token is the one starting with ``m`` (en
+       ``mm``, de ``MM``/Monat, fr ``mm``/mois, es mes, it mese, pt mês, nl
+       maand) while the day letter varies (``dd``, ``TT``, ``jj``, ``gg``) —
+       a per-locale table in the sense the scraping rules require, documented
+       here as such. Locales outside it (non-Latin scripts) fall through.
     """
     numbers = re.findall(r"\d+", current_value)
+    if len(numbers) == 3:
+        values = [int(n) for n in numbers]
+        year_pos = max(range(3), key=lambda i: (len(numbers[i]) >= 4, values[i]))
+        rest = [i for i in range(3) if i != year_pos]
+        a, b = values[rest[0]], values[rest[1]]
+        year = values[year_pos]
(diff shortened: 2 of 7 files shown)


manasnisar:
Addressed the timezone finding in d705815: the past-time guard now only rejects a moment that is already past in **every** timezone (trails UTC by more than the westernmost UTC-12 offset). Anything nearer is passed through for LinkedIn to judge in the profile's timezone — the flow already returns `schedule_rejected` with the picker's own values when LinkedIn refuses. Added a regression test that a moment one hour behind the server clock is no longer rejected locally.

manasnisar:
Addressed the date-reversal finding: the day/month order is now **proven** before anything is filled — the prefill renders LinkedIn's own "today" (within one calendar day of server UTC today), and two distinct day/month readings can never both fall in that three-day window; on day==month prefills the placeholder's m-initial token decides (documented per-locale table). If neither source decides and the target's day ≠ month, the tool returns `schedule_rejected` instead of gambling. The accepted-value check is now position-by-position against the proven order, replacing the unordered set comparison. Verified live with the reversal-shaped target Sep 3 (month 9, day 3): LinkedIn rendered "Posting Thu, Sep 3".

manasnisar:
The UTC+14 finding is a false positive — the arithmetic is inverted. The guard rejects only when the naive wall time trails the UTC wall clock by more than 12 hours (`when <= now_utc - 12h`). At UTC+14, profile-local *now* is the wall time `now_utc + 14h`; one hour in the profile's future is `now_utc + 15h`, which is 15 hours **ahead** of the UTC wall clock, not 13 behind — it clears the cutoff by 27 hours and reaches LinkedIn. Generally, any strictly-future profile-local time at offset k ∈ [-12, +14] is the wall time `now_utc + k + ε > now_utc - 12h`, so no valid future schedule in any timezone can be rejected by this guard. The only boundary it clips is a wall time exactly 12h behind UTC, which is "this very minute" in UTC-12 — not a schedulable future slot. The existing regression test (`test_schedule_post_defers_near_past_to_linkedin`) covers the pass-through direction.

PR #692, opened 2026-08-08
feat(post): Add get_scheduled_posts tool

### LINKED ISSUE #691: [FEATURE] Add get_scheduled_posts tool
## Feature Description

Add a `get_scheduled_posts` tool that lists the authenticated user's scheduled posts (the composer's "View all scheduled posts" modal) as raw text.

## Use Case

Companion reader for `schedule_post` (#689): after scheduling, an assistant needs to see what is queued — scheduled moments and body previews — to verify a schedule landed, avoid double-posting, and pick an entry to modify.

## Suggested Approach

- LinkedIn exposes the list only behind the composer's schedule dialog; the modal replaces the dialog in place and the URL never changes, so the flow is: open composer by URL → clock button → "View all scheduled posts" (the schedule dialog's only button carrying the `arrow-right-small` icon test hook) → read the modal's innerText.
- Read-only in effect: the composer is opened, never typed into, and dismissed empty on every path.
- Returns the repo-standard `{url, sections: {scheduled_posts: raw_text}}`.

manasnisar:
Closes #691

> [!NOTE]
> **Stacked on #690** (`schedule_post`) — review only the last commit here; the first two are #690's. Cross-fork PRs can only target upstream branches, so the base is `main`.

## What

Adds a `get_scheduled_posts` tool that lists the authenticated user's scheduled posts as raw text — the companion reader for `schedule_post`: verify a schedule landed, avoid double-posting, and see what is queued.

Returns the repo-standard `{url, sections: {scheduled_posts: raw_text}}`; entries appear in LinkedIn's own order (soonest first) as "Posting <day>, <date> at <time>" lines followed by body previews.

## How

- LinkedIn exposes the list **only** behind the composer's schedule dialog ("View all scheduled posts"): the modal replaces the dialog in place and the address bar never changes, so there is no URL to navigate to. Flow: composer by URL → clock button → view-all button, identified by its `arrow-right-small` icon test hook (an icon *name*, not a translation).
- Read-only in effect: the composer is opened to reach the list, never typed into, and dismissed empty on every path (the empty-composer dismissal established in #690 — no locale-dependent discard prompt).
- Refactor: the composer/schedule-dialog opening moves out of `schedule_post` into a shared `_open_schedule_dialog`, so both tools ride one measured flow.

## Verification (live)

Called against the live account: returned all scheduled posts (7 entries at the time) with their scheduled moments and full body previews, then left no dialogs behind. `ruff`, `ty`, and the affected test files pass (341 tests).

## Synthetic prompt

> Add a `get_scheduled_posts` tool that reads LinkedIn's "Scheduled posts" modal (composer → clock → "View all scheduled posts", found via its `arrow-right-small` icon test hook — the modal has no URL of its own). Return `{url, sections: {scheduled_posts}}` raw text. Keep it read-only in effect: never type in the composer and dismiss it empty. Factor the composer/schedule-dialog opening shared with `schedule_post` into one helper. Add tool, tests, README/manifest entries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ This MCP server is **free** and **open source**, supported by [**Unipile**](http
 | `get_feed` | Get recent posts from the authenticated user's home feed | working |
 | `search_posts` | Search posts/content globally by keyword (the "Posts" tab) with an optional recency filter (past-24h/past-week/past-month) | working |
 | `schedule_post` | Schedule a feed post via LinkedIn's native Schedule-for-later flow (requires confirmation; publishes even when this server is offline) | working |
+| `get_scheduled_posts` | List the authenticated user's scheduled posts (scheduled moment + body preview per entry) | working |
 | `close_session` | Close browser session and clean up resources | working |
 
 <br/>
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ This MCP server is **free** and **open source**, supported by [**Unipile**](http
 | `get_job_details` | Get detailed information about a specific job posting | working |
 | `get_feed` | Get recent posts from the authenticated user's home feed | working |
 | `search_posts` | Search posts/content globally by keyword (the "Posts" tab) with an optional recency filter (past-24h/past-week/past-month) | working |
+| `schedule_post` | Schedule a feed post via LinkedIn's native Schedule-for-later flow (requires confirmation; publishes even when this server is offline) | working |
 | `close_session` | Close browser session and clean up resources | working |
 
 <br/>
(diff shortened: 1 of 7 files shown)


manasnisar:
Addressed the empty-read finding: an empty `inner_text` from the modal now returns an explicit `list_read_failed` status instead of a section-less result — the modal always carries its own heading text, so emptiness is evidence of a broken read, never of an empty schedule (an empty queue still renders text). Added a regression test that a raising read reports `list_read_failed` with no `sections` key.

PR #695, opened 2026-08-08
feat(post): Add edit_scheduled_post tool

### LINKED ISSUE #693: [FEATURE] Add edit_scheduled_post tool
## Feature Description

Add an `edit_scheduled_post` tool that updates a scheduled post's body text, scheduled moment, or both, via the scheduled-list entry's "Edit post" action.

## Use Case

Completes the scheduling loop with #689/#691: an assistant that schedules and lists posts also needs to fix a typo or move a slot without deleting and re-creating the entry (which would lose its place and history).

## Suggested Approach

- Selection by `match_text` (a snippet of the entry's body), re-resolved against the live list at action time; error on zero or multiple matches, optional `occurrence` disambiguator. Entries expose no stable ids, and a bare index shifts whenever a queued post publishes.
- Entry menu items identified by icon test hooks (`edit-medium`, `clock-medium`, `trash-medium`), never label text; menu items are `div[role=button]`, distinct from native composer buttons.
- "Edit post" opens the full composer with the schedule chip, so one flow covers text, time, or both. In edit mode the composer's last button is a trailing Back, so the save action is the last button *without* an aria-label (attribute presence, locale-safe).
- Same explicit confirmation gating as schedule_post; dry run applies nothing and bails by navigation.

manasnisar:
Closes #693

> [!NOTE]
> **Stacked on #692** (`get_scheduled_posts`), which stacks on #690 (`schedule_post`) — review only the last commit; the earlier ones belong to those PRs. Cross-fork PRs can only target upstream branches, so the base is `main` (GitHub's stacked-PRs linking refused the cross-fork stack).

## What

Adds an `edit_scheduled_post` tool that updates a scheduled post's **body text, scheduled moment, or both, in place** — one tool with optional `new_text` / `new_date` / `new_time` args, gated by `confirm_edit` with a full dry-run mode.

## Selection strategy

The post is identified by `match_text`, a snippet of its current body, **re-resolved against the live list at action time** (with an optional `occurrence` disambiguator and an error on zero/multiple matches). Entries expose no stable ids, and a bare index shifts whenever a queued post publishes — a snippet cannot silently move to a different post. Before any change, the flow additionally verifies the opened composer actually contains `match_text` (`edit_mismatch` otherwise).

## How (measured on the live composer)

- The entry's "…" menu actions are `div[role="button"]` elements identified by **icon test hooks** (`edit-medium`, `trash-medium`, `clock-medium`), never label text; being non-native buttons, the selectors can't collide with composer buttons.
- "Edit post" opens the full composer with the schedule chip, so one flow covers text and time. The schedule dialog's prefills carry the post's *current* schedule and serve as the format example plus the keep-value fallback for whichever component is omitted.
- **Date-order safety carried over from #690's fix, adapted:** the edit prefill does *not* render today, so the today-window order proof is withheld (`prefill_renders_today=False`) — a coincidental window hit on an arbitrary scheduled date could prove the wrong order. The >12 rule and the placeholder table carry the proof; when undecidable and day ≠ month, the tool refuses (`schedule_rejected`) rather than gambling the calendar day. Accepted values are compared position-by-position against the proven order.
- In edit mode the composer appends a trailing **Back** button, so the last-button-primary convention breaks; the save action is the last visible button **without an `aria-label`** (utility buttons are labelled for screen readers, the primary is labelled by its visible text — attribute presence, locale-safe).
- The schedule dialog's confirm is its last *enabled* visible button: Next when something changed, Back when LinkedIn considers it unchanged — both return to the composer correctly.
- **Unsaved exits navigate away** instead of dismissing: a dirty composer's dismiss raises a discard prompt whose buttons only differ by label text. An SPA route change tears the modal down and the stored post keeps its state.

## Verification (live, on a disposable scheduled post)

- Dry run: composer showed the proposed "Posting Wed, Aug 26 at 4:30 PM" + new text (returned in `composer_text`), then the flow abandoned — the scheduled entry was verified **unchanged** via `get_scheduled_posts`.
- Confirmed: entry verified changed to the new text at "Aug 26 at 4:30 PM", original gone.
- Timezone validation mirrors #690's UTC-12 bound (date-only edits are checked at their most generous end-of-day reading; time-only edits are LinkedIn's call).
- Full suite passes: 1837 passed, 22 skipped; `ruff`, `ty`, pre-commit clean.

## Synthetic prompt

> Add an `edit_scheduled_post` tool: select a scheduled post by a body-text snippet re-resolved against the live scheduled-posts list (no stable ids; indices shift), open its "Edit post" menu action (icon test hooks, `div[role=button]`), then change text and/or the scheduled moment in the composer. Withhold the today-window date-order proof for edit prefills (they render the post's schedule, not today), save via the last visible button without an aria-label (edit mode appends a trailing Back), and abandon unsaved changes by navigating away to dodge the locale-texted discard prompt. Gate with `confirm_edit` + dry run; verify live both that dry runs leave the post untouched and that confirmed edits land. Tests, README/manifest entries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ This MCP server is **free** and **open source**, supported by [**Unipile**](http
 | `search_posts` | Search posts/content globally by keyword (the "Posts" tab) with an optional recency filter (past-24h/past-week/past-month) | working |
 | `schedule_post` | Schedule a feed post via LinkedIn's native Schedule-for-later flow (requires confirmation; publishes even when this server is offline) | working |
 | `get_scheduled_posts` | List the authenticated user's scheduled posts (scheduled moment + body preview per entry) | working |
+| `edit_scheduled_post` | Edit a scheduled post's body and/or scheduled moment in place, selected by a body-text snippet (requires confirmation) | working |
 | `close_session` | Close browser session and clean up resources | working |
 
 <br/>
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ This MCP server is **free** and **open source**, supported by [**Unipile**](http
 | `get_feed` | Get recent posts from the authenticated user's home feed | working |
 | `search_posts` | Search posts/content globally by keyword (the "Posts" tab) with an optional recency filter (past-24h/past-week/past-month) | working |
 | `schedule_post` | Schedule a feed post via LinkedIn's native Schedule-for-later flow (requires confirmation; publishes even when this server is offline) | working |
+| `get_scheduled_posts` | List the authenticated user's scheduled posts (scheduled moment + body preview per entry) | working |
 | `close_session` | Close browser session and clean up resources | working |
 
 <br/>
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ This MCP server is **free** and **open source**, supported by [**Unipile**](http
 | `get_job_details` | Get detailed information about a specific job posting | working |
 | `get_feed` | Get recent posts from the authenticated user's home feed | working |
 | `search_posts` | Search posts/content globally by keyword (the "Posts" tab) with an optional recency filter (past-24h/past-week/past-month) | working |
+| `schedule_post` | Schedule a feed post via LinkedIn's native Schedule-for-later flow (requires confirmation; publishes even when this server is offline) | working |
 | `close_session` | Close browser session and clean up resources | working |
 
 <br/>
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -4542,7 +4542,25 @@ class LinkedInExtractor:
                 "list_unavailable",
                 "LinkedIn did not open the scheduled posts view.",
             )
-        await asyncio.sleep(1.0)  # let the list hydrate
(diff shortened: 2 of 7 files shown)


manasnisar:
Addressed the ambiguous-snippet finding: `occurrence` now defaults to **None**, which demands a unique match — several matching entries return `entry_ambiguous` instead of silently picking the first. Passing an integer is an explicit statement that the caller has seen the list (via `get_scheduled_posts`) and is choosing among known duplicates. Added unit tests for unique-match resolution, the ambiguity error, explicit duplicate choice, out-of-range occurrence, and no-match.

manasnisar:
Two further hardenings landed on this branch (also covering the mutable-index finding filed on #696, since the shared resolution lives here):

- **Single-snapshot resolution + pre-click identity check.** Matching is now a pure function over one list snapshot, so the match, preview, and count describe the same moment; because the click is positional against a live queue, a second snapshot taken immediately before it must still show the resolved text at the resolved index, or the action refuses as `entry_moved` instead of landing on whichever post now occupies the position. Unit-tested with two diverging snapshots.
- **Blank snippets refused** at both the tool layer and resolution: normalized to an empty needle, a blank `match_text` was contained in every entry, and with one queued post that "unique" match acted on a post the caller never named.

PR #696, opened 2026-08-08
feat(post): Add delete_scheduled_post tool

### LINKED ISSUE #694: [FEATURE] Add delete_scheduled_post tool
## Feature Description

Add a `delete_scheduled_post` tool that removes a scheduled post via the scheduled-list entry's "Delete post" action.

## Use Case

Completes the scheduling loop with #689/#691: a queued post that is no longer wanted currently requires manual UI work; assistants that schedule (and will edit, see the edit_scheduled_post issue) need the reverse operation, e.g. to clean up test or superseded posts.

## Suggested Approach

- Same `match_text` + optional `occurrence` selection and icon-test-hook menu identification as the edit tool (`trash-medium` item).
- Explicit `confirm_delete` gating; the dry run resolves the entry and reports what would be deleted without opening the confirmation dialog.
- Verify after confirming by re-reading the list and checking the entry is gone.

manasnisar:
Closes #694

> [!NOTE]
> **Stacked on #695** (`edit_scheduled_post`) → #692 → #690 — review only the last commit; the earlier ones belong to those PRs. Cross-fork PRs can only target upstream branches, so the base is `main`.

## What

Adds a `delete_scheduled_post` tool: remove a queued post via LinkedIn's own Delete action, selected by the same live-resolved body-snippet (`match_text` + optional `occurrence`) as `edit_scheduled_post`, gated by `confirm_delete`.

## How

- **Dry run stops before the confirmation surface.** LinkedIn's dialog warns "You won't be able to recover it", so with `confirm_delete=false` the tool only resolves the entry and reports what would be deleted — the menu is never opened.
- Menu action found by its `trash-medium` icon test hook (a `div[role="button"]`, so it cannot collide with native composer buttons).
- **Confirmation dialog identified structurally**: it is the visible dialog holding *no* entry-overflow icons. Measured why the naive alternative fails: DOM order does not put the topmost dialog last — "last visible dialog" resolved to the still-open list modal underneath, whose trailing Back button then failed Playwright's actionability under the confirmation overlay. Its primary is its last button (measured roster: Dismiss, Cancel, Delete).
- **Deletion is verified, not assumed**: after confirming, the entry map is re-read and the snippet must be absent; otherwise the tool reports `delete_unconfirmed`.

## Verification (live)

- Dry run: resolved and previewed the disposable entry ("Posting Wed, Aug 26 at 4:30 PM …"), list unchanged (8 entries).
- Confirmed: returned `deleted`, and the list re-read showed the entry gone (7 entries — which also cleaned up the disposable post used across this PR stack's verification).
- Full suite: 1840 passed, 22 skipped; `ruff`, `ty`, pre-commit clean.

## Synthetic prompt

> Add a `delete_scheduled_post` tool mirroring `edit_scheduled_post`'s snippet-based selection: resolve the entry against the live scheduled-posts list, open its trash-medium menu action, and confirm via LinkedIn's delete dialog — identified as the visible dialog with no entry-overflow icons (topmost is not last in DOM order), primary = last button. Keep the dry run from ever opening the confirmation, and verify deletion by the entry's absence afterwards. Tests, README/manifest entries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ This MCP server is **free** and **open source**, supported by [**Unipile**](http
 | `search_posts` | Search posts/content globally by keyword (the "Posts" tab) with an optional recency filter (past-24h/past-week/past-month) | working |
 | `schedule_post` | Schedule a feed post via LinkedIn's native Schedule-for-later flow (requires confirmation; publishes even when this server is offline) | working |
 | `get_scheduled_posts` | List the authenticated user's scheduled posts (scheduled moment + body preview per entry) | working |
+| `edit_scheduled_post` | Edit a scheduled post's body and/or scheduled moment in place, selected by a body-text snippet (requires confirmation) | working |
 | `close_session` | Close browser session and clean up resources | working |
 
 <br/>
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ This MCP server is **free** and **open source**, supported by [**Unipile**](http
 | `schedule_post` | Schedule a feed post via LinkedIn's native Schedule-for-later flow (requires confirmation; publishes even when this server is offline) | working |
 | `get_scheduled_posts` | List the authenticated user's scheduled posts (scheduled moment + body preview per entry) | working |
 | `edit_scheduled_post` | Edit a scheduled post's body and/or scheduled moment in place, selected by a body-text snippet (requires confirmation) | working |
+| `delete_scheduled_post` | Delete a scheduled post, selected by a body-text snippet (requires confirmation; not recoverable) | working |
 | `close_session` | Close browser session and clean up resources | working |
 
 <br/>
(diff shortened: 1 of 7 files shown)


manasnisar:
Carried the unique-match `occurrence` semantics from #695 into this tool as well: `occurrence` defaults to None and several matching entries error as `entry_ambiguous` instead of silently picking the first — a silent first pick could delete the wrong post, which LinkedIn cannot recover.

manasnisar:
All four findings triaged:

- **Blank selector deletes first post** — fixed: blank `match_text` is refused at the tool layer and again in shared resolution (`entry_not_found` with an explicit message); unit tests cover empty/whitespace snippets.
- **Duplicate snippets fail verification** — fixed: deletion evidence is now the match count dropping by exactly one (the count is captured by the same snapshot that resolved the entry), so surviving duplicates verify as `deleted` instead of inviting a retry that would delete a second post; the unchanged-count case still reports `delete_unconfirmed` and now tells the caller to re-read the list before retrying. Unit tests cover both.
- **Mutable index targets wrong post** — fixed in the shared resolution on #695: single-snapshot matching plus a pre-click identity re-check that refuses as `entry_moved` when the list shifted.
- **Manifest bypasses release ownership** — false positive: CLAUDE.md reserves only the *version* fields of `manifest.json` for the CI release workflow ("the CI release workflow automatically updates manifest.json and docker-compose.yml **with the new version**"). The tools list in that file is maintained by hand — every existing tool is listed there from manual edits, and CLAUDE.md's own PR example says to update "fields, tests, docs, and manifest" when adding capability.

Live re-verification after the restructure: a disposable post went through schedule → edit → delete end-to-end (`scheduled` / `edited` / `deleted`), leaving the queue exactly as it started.

Issue #688, opened 2026-08-07
Windows: ImportError 'DLL load failed while importing _greenlet' with default uvx resolution (greenlet 3.5.4); pinning 3.2.4 fixes it


singhguri:
## Environment
- Windows 10/11 (git-bash), Python 3.11
- Installed via `uvx mcp-server-linkedin@latest` (uv 0.7.x)

## Symptom
The server crashes on startup before doing anything:
```
ImportError: DLL load failed while importing _greenlet: The specified module could not be found.
```
Full traceback: `patchright.async_api` -> `patchright._impl._greenlets` -> `greenlet` -> `_greenlet` C-extension fails to load.

## Root cause
`uvx mcp-server-linkedin@latest` resolves **greenlet 3.5.4** whose Windows wheel's `_greenlet.pyd` fails to import (missing VC runtime dependency / broken wheel for this platform). The same machine works fine with a pinned older greenlet.

## Fix that works
```
uv tool run --with greenlet==3.2.4 --with mcp-server-linkedin mcp-server-linkedin
```
Verified: server starts, all tools (`get_person_profile`, `search_people`, `get_company_profile`, etc.) work, login flow works.

## Suggested fix
- Pin `greenlet<3.3` (or >=3.2.4,<3.3) in the package's dependency spec, or
- Investigate why greenlet 3.5.4's Windows wheel fails to load and bump to a fixed release.

Happy to provide more logs. This blocks all Windows users of the server.

stickerdaniel:
Thanks for the report, the reproduction was precise enough to track this all the way down.

The wheel isn't broken, it gained a system dependency. greenlet moved its Windows builds from Appveyor to GitHub Actions in 3.3.1 and lost the `GREENLET_STATIC_RUNTIME` flag on the way, so `_greenlet.pyd` links the C++ runtime dynamically now:

| release | `_greenlet.cp312-win_amd64.pyd` imports |
| --- | --- |
| 3.2.4 | `KERNEL32.dll`, `python312.dll` |
| 3.3.0 | `KERNEL32.dll`, `python312.dll` |
| 3.3.1 through 3.5.4 | plus `MSVCP140.dll`, `VCRUNTIME140.dll`, `VCRUNTIME140_1.dll` |

[...]

PR #698, opened 2026-08-08
fix(windows): Name the runtime behind a DLL error

### LINKED ISSUE #697: [FEATURE] Name the missing Visual C++ runtime when greenlet cannot load
### Feature Description

On Windows the server can exit before it starts, with nothing that points at why:

```
ImportError: DLL load failed while importing _greenlet: The specified module could not be found.
```

Turn that into a message naming what is missing and where it comes from, and record it in the troubleshooting sections of the README.

### Use Case

Reported in #688, and independently by the same user against another project that ships this server as a channel. Both landed on pinning `greenlet<3.3` as the fix, which is the wrong medicine: the wheel is fine, it gained a system dependency.

greenlet moved its Windows builds from Appveyor to GitHub Actions in 3.3.1 and lost the `GREENLET_STATIC_RUNTIME` flag on the way, so `_greenlet.pyd` links the C++ runtime dynamically now. Import tables of the published `cp312-win_amd64` wheels:

| release | imports |
| --- | --- |
| 3.2.4 | `KERNEL32.dll`, `python312.dll` |
| 3.3.0 | `KERNEL32.dll`, `python312.dll` |
| 3.3.1 through 3.5.4 | plus `MSVCP140.dll`, `VCRUNTIME140.dll`, `VCRUNTIME140_1.dll` |

`MSVCP140.dll` comes with the Visual C++ redistributable and no Python distribution ships it. The python.org installer and the python-build-standalone builds `uv` uses both carry `vcruntime140.dll` and `vcruntime140_1.dll` and stop there, so a `uvx` user is more exposed than most.

We never call greenlet. patchright imports it unconditionally from `_impl/_connection.py`, on the async-only path too, so it loads in every run and a failure there stops the server before anything of ours executes.

Fixed upstream in python-greenlet/greenlet#525 / python-greenlet/greenlet#526, which does not retire this: the affected wheels stay on PyPI for good, and any resolver landing on one reproduces the failure.

### Suggested Approach

A probe in the package `__init__`, which is the only place both entry paths reach before `cli_main` pulls in patchright. Windows only, and free on the path that matters because patchright imports the same module moments later. Match on `DLL load failed`, which CPython formats itself and is English on every install, and re-raise anything else untouched so a merely absent greenlet is not handed advice about a redistributable.

stickerdaniel:
Closes #697

On Windows the server can exit before it starts, with an `ImportError` about `_greenlet` and a DLL that is never named. greenlet moved its Windows builds from Appveyor to GitHub Actions in 3.3.1 and lost the `GREENLET_STATIC_RUNTIME` flag on the way, so `_greenlet.pyd` links the C++ runtime dynamically and needs `MSVCP140.dll`. That DLL comes with the Visual C++ redistributable and no Python distribution ships it, so a machine without the redistributable cannot load greenlet at all. We never call greenlet ourselves; patchright imports it unconditionally on the async-only path, which is why the failure lands before any of our code runs.

The probe lives in the package `__init__` because that is the only place both entry paths reach before `cli_main` pulls in patchright, and it costs nothing on the path that matters since patchright imports the same module moments later. It runs on Windows only. It matches on `DLL load failed`, a prefix CPython formats itself in `dynload_win.c` and keeps English on every install, and re-raises anything else untouched so a merely absent greenlet is not handed advice about a redistributable. The message names both routes, because installing the redistributable needs administrator rights that a `uvx` user may not have, and the stopgap it offers is marked x86-64 only since there are no Windows ARM64 wheels before 3.3.1. Two README troubleshooting sections gain the same pointer.

The upstream fix is filed as python-greenlet/greenlet#525 with the one-line restore in python-greenlet/greenlet#526, where CI now builds x64 and ARM64 wheels that import nothing beyond `KERNEL32.dll` and the Python DLL again. That does not retire this guard: the affected wheels stay on PyPI, and any resolver landing on one reproduces the failure.

Eight tests cover the translated message, the preserved cause, the re-raise, the platform gate and the import ordering, and they run on every platform because the reasoning about a Windows failure should not need a Windows machine; full suite green.

See also: #688

## Synthetic prompt

> On Windows without the Visual C++ redistributable the server dies at import with `DLL load failed while importing _greenlet`, because greenlet's wheels stopped linking the C++ runtime statically in 3.3.1. Add a Windows-only startup probe that turns that bare ImportError into a message naming `MSVCP140.dll` and the redistributable, with a stopgap for machines without admin rights. Put it where both entry paths reach it before patchright is imported, keep the detection locale-independent, re-raise unrelated import failures unchanged, and add README troubleshooting entries plus tests that run on every platform.


stickerdaniel:
* **#698** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/698?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/698?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
### Merge activity

* **Aug 9, 12:46 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/698).
* **Aug 9, 12:46 AM UTC**: @stickerdaniel merged this pull request with [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/698).

PR #700, opened 2026-08-08
test(ci): Run the identity gate on every platform

### LINKED ISSUE #699: [FEATURE] Check the browser identity in CI instead of remembering it
### Feature Description

Nine merged pull requests decided what this browser says about itself, and every one of them was settled by hand: a user agent read off one surface, client hints off another, a frame rate counted against a control window, an outer window compared with the screen under it. None of that is repeatable, and nothing in the suite goes red if a dependency bump undoes it.

That is not a hypothetical. The lock has moved Chromium twice in a week, 148.0.7778.96 to 149.0.7827.55, and both times the only reason anyone knew the identity still held was that someone went and measured it again.

Ask the browser instead of remembering what it answered. A loopback origin serving a page, a dedicated worker and a service worker, each fetching an echo endpoint that returns the headers it saw, and assertions comparing what the browser says about itself in one place against what it says in another.

### Use Case

A contradiction between two surfaces of one browser is the cheapest thing for a site to find, and a version bump produces one by accident. So the guard has to be a test, and it has to run where the merges happen.

Two properties decide whether it is worth having.

**Relations, not values.** Nothing pins a version, a brand string or a screen size. All of those move on their own, and a test that pins them is one somebody deletes after the third false alarm. What cannot move is that the page and its workers agree, that the advertised brand major matches the user agent, and that the window fits the screen it claims to stand on.

**It fails rather than skips.** `browser_dom` skips when no browser is installed, which is right for it: it checks extraction JS, and running it locally is a convenience. A gate that skips itself when the thing it guards is missing is not a gate, so a missing browser or display has to be a CI failure. Locally a skip is still fine, with a reason that names the command to run.

### Suggested Approach

Assert the relations against both launch modes, because one of them is inert on Linux. `hidden_target_is_supported()` is macOS-only, so the default (`headless=True`) is the windowless hidden target on macOS and Chromium's own headless mode on Linux, where it does announce itself. Asserting "no headless token" against the default would therefore assert nothing on the platform CI runs; the headed mode is where it bites, and it is also where the container image is heading.

The page has to collect its own values in a `<script>` and publish them to the DOM, with the test only reading the result: `page.evaluate()` runs in an isolated world, which is not the realm a website gets. Loopback over plain HTTP, because `http://127.0.0.1` is a secure context by definition, so service workers and `navigator.userAgentData` are available without a certificate perturbing the measurement.

Two traps that would make it fail for the wrong reason. The JS channel (`getHighEntropyValues()`) and the header channel (`Sec-CH-UA-Arch` and friends) are separate, and the headers appear only on a request made *after* the server has sent `Accept-CH`. And `platformVersion` is normatively empty on Linux, so asserting it would fail a correct browser.

CI installs no browser today. The existing `test` job needs `patchright install --with-deps chromium --no-shell`, a virtual display for the headed mode, and a cache on `~/.cache/ms-playwright` keyed on the resolved patchright version, or every run pays the download. No new job, so no branch-protection change.

stickerdaniel:
Closes #699

The regression test that ends the hand-measuring landed on this branch first. This is the half that makes it run: CI installs no browser, so the gate would have skipped itself on every pull request while reading as covered.

The `test` job now installs the same browser the server does, `patchright install --with-deps chromium --no-shell`, and starts Xvfb at 1920x1080x24 for the headed mode. The default mode needs no display on Linux, where the hidden target is unsupported and Chromium's own headless mode runs instead, but headed is the mode where the headless token has to be absent, and it is where the container image is heading. The browser is cached on the resolved patchright version, read from the synced environment rather than from `pyproject.toml`, which names a floor and not the version in use. A group mark plus `--dist loadgroup` keeps the identity cases on one xdist worker so they share the two launches they already cache. No new job, so no branch-protection change.

Three more legs run the gate, because on each of them the browser is not the one the Ubuntu job launches. `hidden_target_is_supported()` is macOS-only, so the hidden target, which is the whole reason the default mode can run without announcing itself as headless, is exercised nowhere else; it works on a GitHub runner, which had not been measured before. Linux arm64 is the only platform whose bundle is Playwright's own Chromium rather than Chrome for Testing, and the release publishes that image. Windows has its own binary, its own window manager and its own DPI handling, and one of the things measured is a headed window against the screen it stands on. None of the three blocks a merge: branch protection requires `lint-and-check` and `test` only, which is how `platform-behaviour` has always stood, and the Linux x64 half of the gate sits inside `test` and does block.

Two silent skips came out of running it rather than reasoning about it. The gate assembled its own launch options and so left out `channel="chromium"`; on Linux, where the default mode is physically headless, Playwright resolves the binary from that flag alone and asks for `chromium-headless-shell`, which the setup stopped installing in #669. Thirteen of fifteen cases skipped themselves while the shipped configuration was fine. The gate now launches through `build_launch_options`, the builder both product call sites use. `tests/test_action_signals_dom.py` had the identical gap and now names the channel too; its comment claiming CI installs no browser is no longer true either.

Twelve review rounds turned up rather more than the wiring, and nearly all of it was in the assertions. The harness was missing the cross-origin iframe `docs/browser-fingerprint.md` has always named as the fourth surface, and its service worker answered from a bare async listener the browser is free to terminate at the first await; the three realms were also awaited one after another, so their budgets could add up past the host's and report a merely slow browser as a broken one. Most cases read the default launch alone, leaving a headed-only regression two assertions to get past, and the frame was asked less than the page it was there to be compared with.

The rest were false greens in the assertions themselves. The GREASE filter matched neither spelling in play, so the fake brand counted as a real one behind an `any` that stopped at the agreeing half. The webdriver check read only the prototype, where an own property on `navigator` shadows the accessor; then, once the getter's source was read, it was read with `String()`, which calls the getter's own `toString` and let a JavaScript getter carrying a forged one measure as native. Client hints are structured-field strings, so a blank one is two quote characters and passes a presence check, and the parser smoothed unquoted tokens, unterminated strings, invalid escapes and empty parameters into valid-looking values. The version list, the two brand lists, and the platform and mobile hints were each tied to something and never to each other, so a stale pair of channels, a list naming a different product at the same version, a platform contradicting the user agent, or a desktop machine calling itself mobile all agreed with themselves and passed.

Three things are deliberate and argued where they live rather than left implicit. The GREASE filter names the fake brand instead of deriving it, which is the one value in the file: naming it is what allows every real brand to be required to match the user agent, and the alternative lets a two-brand contradiction through. A patch applied uniformly to one page realm is out of scope, and the reason is recorded with the two escapes that were measured and do not work. And the platform table is fail-open, because an unfamiliar platform is something to look up rather than a reason to fail a browser that is telling the truth.

Verified by running the CI recipe in an Ubuntu container on arm64 and by green CI on all four platforms, each reporting 35 passed with nothing skipped. Every assertion is mutation-checked, twenty-two so far: a bare `BrowserManager` skips thirteen of fifteen; removing the display turns the headed cases red instead of skipping them; dropping the channel from the DOM fixture skips all eleven; a frame reporting a different user agent, different hints or `webdriver` true fails; a webdriver getter redefined plainly, hidden behind an own property, or forged with its own `toString` each fail; the old GREASE pattern fails all four brand cases; a blank, missing, stale or duplicated client hint fails; a full version list naming another product fails; a platform changed on both origins fails; a measurement that raises produces errors rather than skips; a permanently failing launch is attempted once rather than twelve times; a getter redefined only inside the frame fails; a desktop platform claiming a mobile form factor fails; and the group mark lands every case on `gw0`.

## Synthetic prompt

> Wire the browser identity gate into CI. Install the browser the server actually installs, give the headed mode a display, cache it on the resolved patchright version, and keep the identity cases on one xdist worker. Run it on macOS, Windows and Linux arm64 too, since each of those launches a browser the Ubuntu job never does. Make sure the gate goes through the same launch builder the product uses, assert every relation against both launch modes and in every realm including the cross-origin frame the fingerprint doc already asks for, and hold each channel against the other rather than merely requiring a value to be present.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,4 +119,38 @@ jobs:
       - name: Run platform behaviour tests
         run: uv run pytest tests/test_profile_lease.py tests/test_profile_lease_integration.py tests/test_private_state.py tests/test_daemon_lock.py tests/test_greenlet_runtime.py
 
+      # The identity gate belongs here as much as on Ubuntu, and for the same
+      # reason this job exists at all: the mechanism it guards is a different
+      # one per platform. `hidden_target_is_supported()` is macOS-only, so the
+      # hidden target -- the whole reason the default mode can run without
+      # announcing itself as headless -- is never launched by the Ubuntu job.
+      # A regression in it would be invisible everywhere CI looks.
+      #
+      # macOS only. Windows falls back to Chromium's headless mode like Linux
+      # does, so the Ubuntu job already covers that shape, and this job would
+      # pay a second browser download for a third copy of the same answer.
+      - name: Resolve the patchright version
+        if: runner.os == 'macOS'
+        id: patchright
+        run: |
+          version=$(uv run python -c 'from importlib.metadata import version; print(version("patchright"))')
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+      - name: Restore the browser
+        if: runner.os == 'macOS'
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: ~/Library/Caches/ms-playwright
+          key: ms-playwright-${{ runner.os }}-${{ steps.patchright.outputs.version }}
+
+      # No `--with-deps` here: playwright installs system dependencies on
+      # Ubuntu and Windows only, and macOS needs none.
+      - name: Install the browser
+        if: runner.os == 'macOS'
+        run: uv run patchright install chromium --no-shell
+
+      - name: Run the identity gate
+        if: runner.os == 'macOS'
+        run: uv run pytest tests/test_browser_identity.py
+
       - run: uv cache prune --ci
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,20 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, windows-latest]
+        include:
+          - os: macos-latest
+            identity: true
+            browsers: ~/Library/Caches/ms-playwright
+          - os: windows-latest
+            identity: false
+          # Here for the browser rather than for the operating system.
+          # Patchright ships Chrome for Testing everywhere except Linux arm64,
+          # which gets Playwright's own Chromium build, and the release
+          # publishes an arm64 image. Without this leg an identity regression
+          # in that build reaches the container with both other jobs green.
+          - os: ubuntu-24.04-arm
+            identity: true
+            browsers: ~/.cache/ms-playwright
     steps:
       - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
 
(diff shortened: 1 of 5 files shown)


stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

PR #701, opened 2026-08-08
fix(mcpb): Start with the proxy fields left blank

### LINKED ISSUE #678: MCPB: blank optional proxy fields pass literal ${user_config.proxy_server}, server exits on startup
## Summary

On a fresh MCPB install in Claude Desktop, the server fails to start with `Server disconnected` whenever the optional proxy fields are left blank — which is the default for anyone not using a proxy.

Claude Desktop does not substitute `${user_config.*}` placeholders for optional fields that were never filled in. It passes the **literal string** through as the environment variable, so the server receives:

```
PROXY_SERVER=${user_config.proxy_server}
```

`load_config()` treats that literal as a real value and validation aborts startup.

This appears to affect every MCPB user who does not configure a proxy, starting with the v4.20.0 proxy feature.

## Environment

| | |
|---|---|
| Extension version | 4.20.1 (MCPB, unsigned local install) |
| Claude Desktop | 1.25927.0 |
| macOS | 26.5.2 (Apple Silicon) |
| uv | 0.9.28 |

## Actual behaviour

`~/Library/Logs/Claude/mcp-server-MCP Server for LinkedIn.log`:

```
[info] Using MCP server command: <home>/.local/bin/uv
[info] Server started and connected successfully
[info] Message from client: method="initialize" id=0
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File ".../linkedin_mcp_server/__main__.py", line 7, in <module>
    main()
  File ".../linkedin_mcp_server/cli_main.py", line 318, in main
    config = get_config()
  File ".../linkedin_mcp_server/config/__init__.py", line 23, in get_config
    _config = load_config()
  File ".../linkedin_mcp_server/config/loaders.py", line 751, in load_config
    config.validate()
  File ".../linkedin_mcp_server/config/schema.py", line 416, in validate
    self.browser.validate()
  File ".../linkedin_mcp_server/config/schema.py", line 237, in validate
    self._normalize_proxy()
  File ".../linkedin_mcp_server/config/schema.py", line 281, in _normalize_proxy
    raise ConfigurationError(
linkedin_mcp_server.config.schema.ConfigurationError: proxy_server needs a host and an explicit port, for example http://proxy.example:8080.
[info] Server transport closed unexpectedly, this is likely due to the process exiting early.
[error] Server disconnected.
```

Claude Desktop's extension detail pane confirms the unsubstituted values:

```
Environment variables
LINKEDIN_MCP_RUNTIME=mcpb
PROXY_SERVER=${user_config.proxy_server}
PROXY_USERNAME=${user_config.proxy_username}
PROXY_PASSWORD=${user_config.proxy_password}
PROXY_BYPASS=${user_config.proxy_bypass}
```

## Steps to reproduce

1. Install the 4.20.1 `.mcpb` bundle in Claude Desktop.
2. Leave all four proxy fields blank.
3. Start Claude Desktop → `Server disconnected`.

Reduced to a shell one-liner — this reproduces it without Claude Desktop at all:

```console
$ LINKEDIN_MCP_RUNTIME=mcpb PROXY_SERVER='${user_config.proxy_server}' \
    uv run --project <ext-dir> -m linkedin_mcp_server </dev/null
ConfigurationError: proxy_server needs a host and an explicit port, for example http://proxy.example:8080.
```

With the variable unset, it starts correctly:

```console
$ LINKEDIN_MCP_RUNTIME=mcpb uv run --project <ext-dir> -m linkedin_mcp_server </dev/null
INFO  Starting MCP server 'mcp-server-linkedin' with transport 'stdio'
```

## Root cause

`config/loaders.py:293` uses a plain truthiness check, so the literal placeholder counts as "set":

```python
if proxy_server_env := os.environ.get(EnvironmentKeys.PROXY_SERVER):
```

`_normalize_proxy()` (`config/schema.py:261-284`) then takes `"${user_config.proxy_server}"`, finds no `://`, prepends `http://`, and `urlsplit` yields `port is None` → raise.

Note the same class of failure exists for the other three fields via the `proxy_username is set without proxy_server` branch at `schema.py:252-258`, so clearing only `PROXY_SERVER` is not sufficient — all four must be unset.

## Suggested fix

MCPB has no way to omit an env var conditionally, so the guard belongs in the loader. Treating an unsubstituted placeholder as absent would fix all four fields at once:

```python
_PLACEHOLDER = re.compile(r"^\$\{user_config\.[^}]*\}$")

def _env(key: str) -> str | None:
    value = os.environ.get(key)
    if not value or _PLACEHOLDER.match(value.strip()):
        return None
    return value
```

Stripping empty/placeholder values early in `load_config()` would also make the `mcpb` runtime path robust to any future optional `user_config` field.

## Workaround for other users

Delete the four `PROXY_*` variables under **Settings → Extensions → MCP Server for LinkedIn → Advanced options → Environment variables**, keep `LINKEDIN_MCP_RUNTIME=mcpb`, and restart Claude Desktop.

stickerdaniel:
Closes #678

A bundle installed from 4.20.0 onwards does not start unless a proxy is configured. Claude Desktop builds its substitution map from the manifest's `default` values and the answers the user gave; an optional field with neither is absent from that map, so `${user_config.proxy_server}` is handed to the server as if it were an address. Validation finds no port and takes the process down before it can say why.

The quieter half of the same bug hits the people the feature was built for. With only the server filled in, the remaining three placeholders reached Chromium as a username, a password and a bypass list. That does not crash: the browser retries the authentication challenge until the page times out, which the README itself describes as the failure that reads like an expired session.

Every optional field now declares `"default": ""`, which substitutes to nothing and is read as unset. The loader drops a placeholder that reaches it anyway, comparing against one exact string per variable so that a password which happens to read `${user_config.token}` survives, and naming the variables in the log once because a proxy silently skipped means the browser goes out on the real address. A `ConfigurationError` now prints its message to stderr and exits, at startup and at the second validation an interactively chosen transport triggers, instead of arriving as the stack that reads like a crash.

`mcpb validate` cannot catch this class, because the schema knows nothing about substitution. So a test holds the line instead: every `${user_config.X}` named anywhere in `mcp_config` has to resolve to a field that is required or carries a default, no default may itself be a placeholder, no field may be named that does not exist, no field may be declared that nothing reads, and the loader's table has to agree with the manifest string for string. `AGENTS.md` states the rule alongside it.

Verified by reproducing the reporter's one-liner against the fix, by replaying Claude Desktop's substitution routine over the new manifest for the blank, server-only and full-credential cases, against `mcpb validate` at the version the release workflow pins, and by mutating each guard to confirm the tests fail without it.

## Synthetic prompt

> Issue #678: MCPB bundles fail to start because Claude Desktop passes the literal `${user_config.proxy_server}` for optional fields left blank. Find out how the host actually resolves those placeholders, fix it at the source and in the loader, and add a test that makes this class of manifest mistake impossible to ship again.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -120,7 +120,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
   and the loader reads an empty variable as unset. `required: true` is the only
   other safe shape, because a host skips the whole MCP config while a required
   field is empty. `tests/test_manifest.py` holds this line; `mcpb validate`
-  does not, and cannot — the schema knows nothing about substitution.
+  does not, and cannot: the schema knows nothing about substitution.
 - **A placeholder that does reach the process is not a value.** `_env()` in
   `config/loaders.py` drops it. Both directions matter and only one is loud:
   `PROXY_SERVER` fails validation and stops the server, while
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -110,6 +110,23 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
   evidence here: its `title` key dates from patchright 1.58.0 and omits the
   `Google` the binary prints. See `_COMPARABLE_PRODUCTS` for the measurements.
 
+## Extension Bundle Rules
+
+- **An optional `user_config` field needs a `default`.** A host substitutes
+  `${user_config.NAME}` from the manifest's defaults plus the answers the user
+  gave; a field in neither is not in that map, so the placeholder is handed to
+  the server verbatim as if it were a setting. Measured in Claude Desktop's own
+  substitution routine. `"default": ""` is enough: it substitutes to nothing,
+  and the loader reads an empty variable as unset. `required: true` is the only
+  other safe shape, because a host skips the whole MCP config while a required
+  field is empty. `tests/test_manifest.py` holds this line; `mcpb validate`
+  does not, and cannot — the schema knows nothing about substitution.
+- **A placeholder that does reach the process is not a value.** `_env()` in
+  `config/loaders.py` drops it. Both directions matter and only one is loud:
+  `PROXY_SERVER` fails validation and stops the server, while
+  `PROXY_USERNAME` is offered to the proxy as a credential and comes back as a
+  timeout that reads like an expired session.
+
 ## Tool Return Format
 
 All scraping tools return: `{url, sections: {name: raw_text}}`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,11 +116,18 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
   `${user_config.NAME}` from the manifest's defaults plus the answers the user
   gave; a field in neither is not in that map, so the placeholder is handed to
   the server verbatim as if it were a setting. Measured in Claude Desktop's own
-  substitution routine. `"default": ""` is enough: it substitutes to nothing,
-  and the loader reads an empty variable as unset. `required: true` is the only
-  other safe shape, because a host skips the whole MCP config while a required
-  field is empty. `tests/test_manifest.py` holds this line; `mcpb validate`
-  does not, and cannot: the schema knows nothing about substitution.
+  substitution routine. `required: true` is the other safe shape, because a
+  host skips the whole MCP config while a required field is empty.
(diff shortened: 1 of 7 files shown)


stickerdaniel:
@greptileai review

PR #705, opened 2026-08-09
fix(docker): Run Chromium under Xvfb

### LINKED ISSUE #704: [FEATURE] Run Docker Chromium headed under Xvfb
### Feature Description

Run the Docker browser headed on a virtual display, while keeping it invisible to the host.

The image currently installs the full browser and the headless shell, leaves `HEADLESS=true`, and starts Python directly under tini. That makes Docker the one shipped runtime which still uses Chromium's real headless mode and says `HeadlessChrome` in its user agent and brand list. It also loses the browser properties the desktop path now guards: plugins, `window.chrome`, and coherent headed geometry.

The candidate has been measured on both published architectures. Headed Chromium under Xvfb at 1920×1080×24 reports no headless token, all four identity realms agree, plugins and `window.chrome` are present, native architecture/bitness are correct (`x86/64` and `arm/64`), and the outer window fits the screen.

### Use Case

The same profile should be opened by one coherent browser everywhere it ships. Docker should not identify itself as a different, stripped browser solely because there is no physical monitor attached.

WebGL is part of that coherence. Ten cold launches on each architecture established the working configuration:

- WebGL1: 10/10
- WebGL2: 10/10
- renderer, byte-identical across all twenty launches: `ANGLE (Mesa/X.org, llvmpipe (LLVM 15.0.6 128 bits), OpenGL 4.5)`
- no SwiftShader and no GPU-process crash

The required switches are `--enable-webgl --ignore-gpu-blocklist`. `--use-gl=angle --use-angle=gl`, `--use-gl=egl`, desktop GL and Vulkan all produced no context. SwiftShader produced a context but is excluded deliberately: Patchright strips that fallback because the renderer string is itself a signal.

### Suggested Approach

- Install only full Chromium with `patchright install chromium --no-shell`.
- Add `xauth` beside `tini`; Patchright's dependency install supplies Xvfb but omits xauth under `--no-install-recommends`.
- Default `HEADLESS=false`, still overridable.
- Start Xvfb directly in a small entrypoint, wait for its socket, then `exec` Python. Run the entrypoint under `tini -g` so TERM reaches both Python and Xvfb.
- Add the two measured WebGL switches to the browser launch arguments in container runtimes only.
- Refuse the experimental shared-browser daemon in a container. Its owner starts a new session and outlives the stdio frontend, while Xvfb belongs to that frontend's process group.

`xvfb-run` is not a correct entrypoint here. Measured under `docker stop`, `tini -g -- xvfb-run ... python` exited 143 without Python's shutdown path: `xvfb-run` handles its own EXIT trap but does not forward TERM to its child. Starting Xvfb in the same process group and replacing the wrapper with Python produced Uvicorn's complete graceful-shutdown sequence before container exit.

The container still has no visible login UI. It keeps using the documented host-side login until the separate viewer work ships.

stickerdaniel:
Closes #704

Docker was the one shipped runtime still using Chromium's real headless mode. The image installed full Chromium and the headless shell, left `HEADLESS=true`, and started Python with no display. That made the same saved session identify as a stripped browser in Docker: `HeadlessChrome` in the user agent and brand list, no plugins, no `window.chrome`, and no WebGL context.

The image now runs full Chromium headed on Xvfb at 1920×1080×24. Nothing appears on the host. `HEADLESS=false` is an image default and remains overridable; `true` deliberately returns to Chromium's headless mode. The `--no-shell` install reduces the uncompressed image from 543.5 to 429.6 MiB on arm64 (-113.9 MiB), and from 512.4 to 406.2 MiB on amd64 (-106.2 MiB), in clean builds of the same source.

WebGL was measured across the available renderer configurations. Headed Chromium under Xvfb produced no WebGL1 or WebGL2 context on either architecture, despite Mesa's software rasteriser being present. Explicit ANGLE GL, EGL, desktop GL and Vulkan selectors all stayed there. `--enable-webgl --ignore-gpu-blocklist` lets Chromium use the existing Mesa path without choosing one: ten cold launches on each architecture produced WebGL1 and WebGL2 every time, with one byte-identical renderer across all twenty runs, `ANGLE (Mesa/X.org, llvmpipe (LLVM 15.0.6 128 bits), OpenGL 4.5)`. No SwiftShader and no GPU-process crash. The same runs kept every identity relation green: no headless token, native `x86/64` and `arm/64` hints, all four realms agreeing, plugins and `window.chrome` present, and the window fitting its screen.

The process topology is part of the fix. `tini -g -- xvfb-run ... python` failed the signal test: `docker stop` exited 143 without Python's shutdown path because `xvfb-run` has an EXIT trap for Xvfb and no TERM forwarder for its child. A small entrypoint starts Xvfb directly, waits for its socket, then supervises it alongside Python. It delivers TERM to both children, waits for Python's graceful shutdown, cleans up Xvfb when Python exits, and fails the container when the display dies first. `tini -g` signals the whole process group, and `-e 143` maps the expected handled-SIGTERM status to success. In the final amd64 and arm64 images, Uvicorn logged `Application shutdown complete`; both containers exited 0 within two seconds.

The experimental shared-browser daemon is ignored in Docker. Its owner deliberately starts a new session so it can outlive a stdio frontend; measured in the candidate, the owner remained separate while Xvfb belonged to the frontend's process group. The owner would lose its display when the frontend exits. The container keeps one browser per frontend.

The final images were built and exercised on both architectures. The production launch path passed ten cold WebGL and identity runs per architecture, including the `DISPLAY=:99.0` screen suffix. After a `SIGKILL`, each architecture restarted successfully in the same container writable layer, then shut down gracefully; host-prefixed displays fail immediately with a precise diagnostic. The updated commit passed every CI check on Linux amd64, macOS, Windows and Linux arm64. Locally, 1,915 tests passed and 11 platform-dependent tests skipped; one daemon-turnover timing bound failed under a load average above 180 and reproduced on a detached `origin/main` checkout. Targeted Docker, launch and daemon tests passed 43/43 with the Bash-5-only supervisor case skipped on macOS Bash 3. Ruff, formatting, Ty, shell syntax and pre-commit checks are clean.

The login flow is unchanged. The container has a virtual display but no viewer for its form, 2FA, or captcha, so profiles are still created on the host until the separate login-viewer work ships.

## Synthetic prompt

> Run Docker Chromium headed on an invisible Xvfb display. Install only the full browser, preserve coherent geometry and Mesa llvmpipe WebGL on amd64 and arm64, deliver `docker stop` to both Python and Xvfb without `xvfb-run`, and refuse the shared-browser daemon whose owner cannot share the display lifetime. Update tests and Docker documentation with the measured results.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -9,8 +9,9 @@
 # session from this server; otherwise start once with --claim-profile-root.
 USER_DATA_DIR=~/.linkedin-mcp/profile
 
-# Browser mode (default: true)
-# true = headless, false = visible window
+# Browser mode (default: true outside Docker; the image overrides it to false)
+# true = Chromium's headless mode; false = headed. Docker's headed window lives
+# on a virtual display and never appears on the host.
 HEADLESS=true
 
 # Logging level (default: WARNING)
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,14 +23,27 @@ ENV PATH="/app/.venv/bin:$PATH"
 ENV PLAYWRIGHT_BROWSERS_PATH=/opt/patchright
 
 RUN patchright install-deps chromium && \
-    patchright install chromium && \
+    patchright install chromium --no-shell && \
     apt-get update && apt-get install -y --no-install-recommends tini && \
     chmod -R 755 /opt/patchright && \
     rm -rf /var/lib/apt/lists/*
 
+COPY --chmod=755 docker-entrypoint.sh /usr/local/bin/linkedin-mcp-entrypoint
+
+# A full headed browser on a virtual display. No window reaches the host, and
+# HEADLESS stays overridable for anyone who deliberately wants Chromium's real
+# headless mode. DISPLAY is fixed inside one container, where there is only one
+# X server to collide with.
+ENV DISPLAY=:99
+ENV HEADLESS=false
+
 USER pwuser
 
-# tini reaps the chromium subprocesses Patchright orphans onto PID 1; without an
-# init the container leaks zombie chrome-headless processes over its lifetime.
-ENTRYPOINT ["tini", "--", "python", "-m", "linkedin_mcp_server"]
+# -g sends TERM to the whole process group: Python gets to run FastMCP's
+# graceful shutdown, and Xvfb leaves with it. The entrypoint supervises both in
+# that group, so either one dying terminates the other instead of leaving a live
+# server with no display. A handled SIGTERM can still report 143, so -e maps
+# that expected `docker stop` path to success. Without an init, Chromium
+# subprocesses orphaned onto PID 1 would also accumulate as zombies.
+ENTRYPOINT ["tini", "-g", "-e", "143", "--", "linkedin-mcp-entrypoint", "python", "-m", "linkedin_mcp_server"]
 CMD []
--- a/README.md
+++ b/README.md
@@ -303,7 +303,7 @@ On startup, the MCP Bundle starts preparing the shared Patchright Chromium brows
 
 ### Authentication
 
-Docker runs headless (no browser window), so you need to create a browser profile locally first and mount it into the container.
+Docker runs full Chromium headed on a virtual display. No browser window reaches the host, so you still need to create a browser profile locally first and mount it into the container.
 
 **Step 1: Create profile on the host (one-time setup)**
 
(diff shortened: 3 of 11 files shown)


stickerdaniel:
* **#705** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/705?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/705?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
### Merge activity

* **Aug 9, 2:37 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/705).
* **Aug 9, 2:37 PM UTC**: @stickerdaniel merged this pull request with [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/705).

PR #715, opened 2026-08-10
fix(bootstrap): self-heal missing Patchright Chromium install

d33disc:
## Summary
- Auto-installs the managed Patchright Chromium browser into the shared cache (`~/.linkedin-mcp/patchright-browsers`) whenever it's missing, instead of hard-failing with "Executable doesn't exist" and requiring a manual `patchright install chromium`.
- Self-heal is wired into three places: the CLI entrypoint (`main()`, before `--login`/`--status` dispatch), the background startup hook, and the tool-call readiness gate (`ensure_tool_ready_or_raise`) -- falls back to the original manual-install error message only if the automatic install itself fails.
- Fixes a test-isolation bug found along the way: `cli_main.py` imports `ensure_browser_installed` by name, a separate binding from `bootstrap.ensure_browser_installed`, so tests exercising `main()` were shelling out to a real network install. Added an autouse `conftest.py` fixture stubbing both import sites.

Note: this only fixes the browser-binary provisioning gap. A dead/expired LinkedIn session (cookie replay rejected, fingerprint mismatch, LinkedIn checkpoint) still requires `--login` -- that's a server-side decision on LinkedIn's end, not something client code should paper over.

## Test plan
- [x] `uv run pytest -q` -- 268 passed (263 baseline + 5 new)
- [x] `uv run ruff check` / `ruff format` clean on all changed files (pre-existing violations in untouched lines left as-is)
- [x] `uv run ty check` clean on changed files
- [x] Reproduced the original failure locally (cleared cache -> `Executable doesn't exist`) and confirmed `--login` now self-heals it


d33disc:
Closing -- this was opened against the wrong repo by mistake (meant to target my own fork, not upstream). Re-opening there.

PR #716, opened 2026-08-11
fix(windows): Pin the greenlet that links its runtime

### LINKED ISSUE #688: Windows: ImportError 'DLL load failed while importing _greenlet' with default uvx resolution (greenlet 3.5.4); pinning 3.2.4 fixes it

stickerdaniel:
Closes #688

greenlet's Windows wheels stopped carrying the C++ runtime inside the extension when its builds moved from Appveyor to GitHub Actions and `GREENLET_STATIC_RUNTIME` was dropped along the way. From 3.3.1 the published `_greenlet.pyd` imports `MSVCP140.dll`, which ships with the Microsoft Visual C++ Redistributable and with neither the python.org installer nor the python-build-standalone builds `uv` installs. patchright imports greenlet unconditionally, so on a Windows machine without that redistributable the server died before its entry point ran.

python-greenlet/greenlet#526 restored the variable and released as 3.5.5. This lock still carried 3.5.1, and the MCPB bundle starts through `uv run --project`, so every bundle kept installing an extension that needs a DLL the machine may not have. Raising the pin removes the problem for that path; `uvx` resolves fresh and already gets 3.5.5. The startup guard from #698 stays as it is: the affected wheels remain on PyPI for good, a greenlet built from source can need that DLL at any version, and nothing forces a third-party install to honour this lock.

Verified by reading the PE import table of the published wheels: 3.5.1 imports `MSVCP140.dll`, `VCRUNTIME140.dll` and `VCRUNTIME140_1.dll`, while 3.5.5 imports `python312.dll` and `KERNEL32.dll` alone, on `win_amd64` and `win_arm64` both.

## Synthetic prompt

> greenlet 3.5.5 shipped the upstream fix for the Windows wheels that need MSVCP140.dll. Raise our lock to it, and update the README troubleshooting entries and the greenlet_runtime docstring to name the released fix instead of the open PR. Keep the startup guard.

Generated with Claude Opus 5 high


stickerdaniel:
@greptileai review

Issue #441, opened 2026-05-10
[BUG] Multi-paragraph message sending resolves to multi-message


frgfm:
### Setup

uvx on macOS

### What Happened

I was running some tests, and using the `send_message` tool of the MCP. I happened to try that out with a two-paragraph message and it sent 2 consecutive messages instead of one.

I thought it was a problem with the linebreak characters so I tried `\n\n`, `\n`, `\n \n` and none worked. Weird thing is that I asked it to parse how it's encoded on messages I sent manually that actually have multiple paragraphs and they were using those characters `\n \n`.

### Steps to Reproduce
Prompt sent to codex/claude
```
Using Linkedin MCP server, send a message to https://www.linkedin.com/in/stickerdaniel/ with this copy "First paragraph\n\nSecond paragraph"
```

### Logs

```shell

```

PR #720, opened 2026-08-12
fix(messaging): confirm real delivery, keep a message whole, clear the composer

### LINKED ISSUE #573: [BUG] send_message can over-confirm sends and miss share-info prompt
## Setup

- Repository branch: local development against `upstream/main`
- Tool: `send_message`
- Runtime/client: MCP clients using the browser-backed LinkedIn session

## What Happened

`send_message` can report `sent: true` by finding the outgoing message text anywhere on the page, including text that is still present in the editable composer or was already present in the visible conversation UI before the send attempt completed.

Some LinkedIn message sends can also show a post-Send prompt asking whether to share profile/contact information with the recipient. The current flow has no explicit handling for that prompt, so it can leave the send unconfirmed or make the final state ambiguous. The safer default after the operator has approved only the message body is to avoid sharing additional profile/contact information.

## Steps to Reproduce

1. Open a message compose flow via `send_message` with `confirm_send=true`.
2. Use message text that is already visible in the page or remains in the composer while LinkedIn is still processing the send.
3. Trigger a LinkedIn post-Send prompt asking whether to share profile/contact information, where available.
4. Observe that the tool can treat page text presence as confirmation instead of requiring a new non-composer occurrence after the send attempt, and it does not explicitly choose the no-share path for the profile/contact-info prompt.

## Logs

No logs with private LinkedIn content are attached. The issue is based on the existing `send_message` confirmation path and reproduced with unit-level tests to avoid sending additional live LinkedIn messages during development.

rcnazar:
Fixes #573, #441, #433.

Three defects in `send_message` that compound. A three-paragraph message goes out as two messages, the third paragraph never leaves the composer, and the tool still returns `"Message sent."`.

### Delivery confirmation (#573)

`_message_text_visible()` checked whether `document.body.innerText` contains the message. The composer is `contenteditable`, so the draft we just typed is part of that text. The check passes before the send button is clicked and cannot fail, which turns every failed send into a reported success.

Now the occurrences are counted outside the composer, and confirmation requires one more than the baseline captured just before sending. That also stops a resend from being confirmed by the copy already sitting in the thread.

### Multi-paragraph messages (#441)

`keyboard.type()` translates `\n` into an Enter press, and Enter is how the composer submits. It is the same gesture this code already uses as its send fallback. So each blank line sent whatever was typed so far, and the final paragraph was left behind in the composer.

Now it types line by line and uses Shift+Enter for the breaks, so the whole text lands as one message.

### Stale composer (#433)

The success path returned without dismissing the composer, leaving it mounted and pointed at the previous recipient. That is what makes the second send of a session fail with `composer_unavailable`. It now dismisses after a send, and again on entry, to cover a call that died mid-flight.

### Tests

12 new tests in `tests/test_scraping.py` covering all three, including the case that matters most: a draft still in the composer must not count as delivery. Two existing tests were updated for the renamed helper. Full suite passes (1930 passed, 58 skipped).

Also verified live against a real 1st-degree conversation. Before the fix, a three-paragraph message arrived as two messages with the signature missing. After, it arrives as a single message with all three paragraphs and accented characters intact.

### Note

The three fixes are in the same code path and the interaction between them is the actual user-visible bug, so they are together here. Happy to split into three PRs if you prefer.

There are older PRs open against some of this (#574, #482). I did not build on them, since they had drifted from main. Close this if either one is closer to what you want.

### DIFF
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -2739,20 +2739,62 @@ class LinkedInExtractor:
         )
         return bool(matched)
 
-    async def _message_text_visible(self, message: str) -> bool:
-        """Wait until the compose page visibly contains the just-sent message text.
+    # Counts occurrences of the message text outside the composer. The composer
+    # is contenteditable, so its own draft text is part of document.body.innerText:
+    # a plain `bodyText.includes(message)` is satisfied by the text we just typed
+    # and reports a send that never happened. Only a new occurrence *outside* the
+    # composer means LinkedIn actually rendered the message into the thread.
+    _MESSAGE_OCCURRENCES_JS = """({ expected }) => {
+        const normalize = value => (value || '').replace(/\\s+/g, ' ').trim();
+        const needle = normalize(expected);
+        if (!needle) return 0;
+        const count = haystack => {
+            let seen = 0;
+            let at = 0;
+            while ((at = haystack.indexOf(needle, at)) !== -1) {
+                seen += 1;
+                at += needle.length;
+            }
+            return seen;
+        };
+        const bodyText = normalize(document.body?.innerText || '');
+        let composerText = '';
+        for (const el of document.querySelectorAll(
+            'div[role="textbox"][contenteditable="true"]'
+        )) {
+            composerText += ' ' + normalize(el.innerText || '');
+        }
+        return Math.max(0, count(bodyText) - count(normalize(composerText)));
+    }"""
+
+    async def _count_message_occurrences(self, message: str) -> int:
+        """Count occurrences of the message in the thread, ignoring the composer."""
+        try:
+            return int(
+                await self._page.evaluate(
+                    self._MESSAGE_OCCURRENCES_JS, {"expected": message}
+                )
+            )
+        except Exception:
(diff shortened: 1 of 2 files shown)


rcnazar:
Both review points were real, fixed in 84e72ec.

The selector one is the worse of the two. `_resolve_message_compose_box` also accepts `main [contenteditable="true"][aria-label*="message"]`, which has no `role="textbox"`, so a composer of that shape was never subtracted and its draft counted as thread text. That is the original false "sent" coming back through a side door. Checked it in a browser against each accepted shape: with the old hardcoded selector that case counts 1 with nothing sent, with the selector list it counts 0. The list is passed in from `_MESSAGING_COMPOSE_FALLBACK_SELECTORS` now so the two cannot drift apart, and nested composers are de-duplicated so shared text is not subtracted twice.

On the baseline, agreed that 0 and "could not read the page" should not be the same value. `_count_message_occurrences` returns `None` for the second case, and an unverifiable send is reported as unconfirmed with a message that says so. Reporting a send that did happen as unconfirmed is recoverable, the reverse is what this PR exists to stop.

On the tests: the Python approximation is there to pin the contract, and the JS itself is exercised in a browser separately. Happy to add a Playwright-backed test in-repo if you want that covered by CI.

PR #722, opened 2026-08-12
fix(search): location filter is silently ignored — emit geoUrn and validate the URN

BasitS-hash:
## The bug

`search_people` accepts a plain-text `location` and appends it as `&location=<text>`:

```python
params = f"keywords={quote_plus(keywords)}"
if location:
    params += f"&location={quote_plus(location)}"      # no validation
if network:
    params += f"&network={_encode_list_facet(network)}"
if current_company:
    params += f"&currentCompany={_encode_list_facet([current_company])}"
```

LinkedIn's people-search backend has no `location` facet. The geo facet is **`geoUrn`**, a JSON-list of numeric geo URN ids. The URL is well-formed, LinkedIn accepts it, and returns the **completely unfiltered** result set. `grep -rn "geoUrn"` across the repo currently returns nothing.

## Why it matters

The failure is silent. Searching `location="Washington DC-Baltimore Area"` returns people in Philadelphia, New Hampshire, North Carolina, Colorado, Missouri and Ohio — with nothing signalling that the filter did not apply. A caller reasonably reads those results as location-filtered and acts on them.

This is the **same defect class already handled for `current_company`**, whose guard and docstring state that plain-text company names are silently ignored by LinkedIn and must be numeric URN ids. `location` was simply left on the old path.

## The fix

Mirrors the existing `currentCompany` treatment exactly:

- emit `geoUrn=` via `_encode_list_facet` instead of a bare `location=`
- raise `FilterValidationError` on non-numeric input, pointing callers at `keywords` as the fallback when they don't have a URN
- update both docstrings (extractor + tool) to describe real behaviour

## Test changes

`test_search_people_combines_all_filters` used `location="Seattle"` and asserted `location=Seattle` appeared in the URL — it was asserting the broken behaviour. It now uses a geo URN and asserts the `geoUrn` facet.

Four regression tests added, mirroring the `currentCompany` set:

| test | asserts |
|---|---|
| `..._location_filter_emits_geo_urn` | `geoUrn=%5B%22103644278%22%5D` present, `&location=` absent |
| `..._rejects_plain_location_name` | `"New York"` raises |
| `..._rejects_unicode_digit_location` | Unicode digits raise (`str.isdigit()` would accept) |
| `..._empty_location_is_noop` | empty string adds no facet |

## Verification

```
uv run pytest -q
1925 passed, 58 skipped
```

## Note on the breaking change

Callers currently passing plain place names will now get a `FilterValidationError` instead of silently-unfiltered results. That seemed right given the precedent `current_company` sets — a loud error beats a filter that quietly does nothing — but happy to switch it to a warning that drops the facet if you'd prefer a softer migration.

### DIFF
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -3613,7 +3613,11 @@ class LinkedInExtractor:
 
         Args:
             keywords: Free-text query ("software engineer", "recruiter at Google").
-            location: Optional location filter ("New York", "Remote").
+            location: Optional location filter. LinkedIn's people-search
+                ``geoUrn`` facet only filters on the numeric geo URN id
+                (e.g. ``"103644278"`` for the United States); plain-text place
+                names are accepted by the URL but ignored by LinkedIn and
+                return the unfiltered result set.
             network: Optional connection-degree filter. Each element is one of
                 ``"F"`` (1st-degree), ``"S"`` (2nd-degree), ``"O"`` (3rd-degree
                 and beyond). Example: ``["F"]`` to only return 1st-degree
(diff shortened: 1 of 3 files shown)


stickerdaniel:
Thanks @BasitS-hash, I'd like to merge this! Could you resolve the conflicts by rebasing onto current `main` and re-verify the location filter against the live LinkedIn DOM?

PR #728, opened 2026-08-15
build(docker): Pin the image build backend

### LINKED ISSUE #655: [FEATURE] Pin the build backend for the Docker image too
### Feature Description

#654 pins the build backend for the published distributions: `uv build` runs with a hashed constraint file, so a tampered `setuptools` release cannot reach the artifacts that go to PyPI.

The Docker image is not covered by that. `Dockerfile:11` runs `uv sync --frozen`, which installs the project itself and therefore invokes `setuptools.build_meta` again. `uv.lock` does not record `[build-system].requires`, so `--frozen` pins nothing here, and `uv sync` has no `--build-constraint` flag to pass the hashed file through.

### Use Case

The Docker job holds the Docker Hub credentials, so a compromised build backend picked up during the image build could end up in the image that gets pushed under the official tag. The distributions are reproducible after #654; the image is not.

### Suggested Approach

Two options, both worth weighing before implementing:

- Install the already-built, hash-verified wheel in the image instead of rebuilding the project with `uv sync`. This reuses exactly the bytes that were published and removes the second build entirely, but it changes how the image is assembled and needs the wheel passed into the build context.
- Add `[tool.uv] build-constraint-dependencies` to `pyproject.toml`, which applies to every uv operation in the project including `uv sync`. Renovate's `pep621` manager does not extract that key, so it would need a custom manager to avoid drifting away from `build-constraints.txt`, and the setting travels into the published sdist.

stickerdaniel:
Closes #655

#654 pinned the PEP 517 build backend for the published distributions: `uv build` runs against a hashed constraint file, so a tampered setuptools release cannot reach PyPI. The image was not covered. `uv sync --frozen --no-dev` installs the project and therefore invokes `setuptools.build_meta` again, `uv.lock` records no `[build-system].requires`, and `uv sync` accepts no build constraint, so setuptools resolved fresh from PyPI inside the one release job that holds the Docker Hub credentials.

The dependency sync keeps `--no-install-project` and the project is installed on its own with `uv pip install --no-deps --build-constraints build-constraints.txt .`, against the same hashed file the distributions use. `uv pip install` takes the constraint that `uv sync` has no flag for, and `--no-deps` leaves the runtime dependency set exactly as the frozen lock installed it. `--require-hashes` is not usable here, because the local directory requirement carries no hash of its own, and it is not needed: uv verifies the hashes a constraint file does carry. What it does not verify is a transitive build dependency with no entry of its own, so the constraint file has to keep carrying the full closure; `build-constraints.in` is what `uv pip compile` reads to produce it, and the tests hold `[build-system].requires` to it.

The issue weighed two other routes. Installing the published wheel would need it passed into the build context and would break a plain `docker build .` for contributors. `[tool.uv] build-constraint-dependencies` takes PEP 508 requirements, which have no hash syntax, so it pins a version without verifying the bytes; it would also need a custom Renovate manager and travels into the sdist. The compiled file stays with Renovate's `pip-compile` manager as it is.

`.dockerignore` now excludes `**/.env*` rather than `.env` alone, along with `.debug`, `CLAUDE.local.md`, `.agents` and `.gemini`. `COPY . .` keeps the whole context in the builder layer, which the release exports through `cache-to: type=gha,mode=max`, and a local `.env.previous-proxy` carries a session cookie and proxy credentials while 9.7 MB of working notes carry nothing the build reads. No published image was affected: the runtime stage takes only `/app/.venv`.

The tests measure this rather than reading it out of the Dockerfile, after several rounds of the latter passing while the build did something else. The builder stage is built and the wheel it produced is read for the setuptools that made it, then rebuilt with every constraint hash corrupted, which has to fail: the version alone proves nothing once the pin equals the current release, which is the state Renovate maintains. Which paths reach the context is asked of BuildKit through a throwaway probe against a synthetic tree, because no reimplementation of `.dockerignore` matching here agreed with it. What a build cannot show stays a string check: that hash verification was not switched off, and that the two constraint files still agree.

Verified by building both stages on `linux/amd64` and `linux/arm64`: the wheel records `setuptools 83.0.0` against `84.0.0` on `main`, corrupted hashes fail the build with `Hash mismatch`, and the runtime `/app/.venv` is byte-identical to the baseline apart from the backend version, the console-script shebang and uv's cache timestamp. Full suite 1973 passing.

The docker-backed tests fail in CI rather than skipping, the way the browser identity gate already does: they are the only checks here that observe the build instead of reading the Dockerfile, and a skipped gate is indistinguishable from a passing one. Enabling that surfaced an unrelated defect this branch was triggering, so one commit here is outside #655 and would otherwise have been filed on its own. `debug_trace` resolves the trace directory once and caches it for the process, and it is derived from the profile root, so a test pointing `USER_DATA_DIR` at its own `tmp_path` left every later test on that xdist worker writing into a directory pytest had removed. A daemon test's real `server.log` then landed where a diagnostics test asserts none exists. `reset_singletons` already resets six other modules for that reason; the trace directory joins them. Adding a docker build is what shifted the distribution enough to put those two tests on one worker.

## Synthetic prompt

> The Docker image rebuilds the project with an unconstrained setuptools, so #654's hashed build backend does not cover it. Install the project separately with `uv pip install --build-constraints`, and cover it with tests that fail when the pin, the flag, the constraint file's reach or the hash verification is removed. Extend `.dockerignore` so no environment file or local working directory reaches the build context.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -165,3 +165,11 @@ assets/*
 .claude
 .github
 .docker
+.agents
+.gemini
+
+# Local working notes and private context. None of it is a build input, and
+# `COPY . .` otherwise keeps all of it in the builder layer, which the release
+# exports through `cache-to: type=gha,mode=max`.
+.debug
+CLAUDE.local.md
--- a/.dockerignore
+++ b/.dockerignore
@@ -103,8 +103,12 @@ celerybeat.pid
 # SageMath parsed files
 *.sage.py
 
-# Environments
-.env
+# Environments. The pattern has to cover the suffixed variants and every
+# directory, because `COPY . .` keeps whatever the context carries in the
+# builder layer, and a `.env.local` holds the same session cookie and proxy
+# credentials as the plain file. That layer is not published, the runtime stage
+# takes only `/app/.venv`, but it is exported by `cache-to: type=gha,mode=max`.
+**/.env*
 .venv
 env/
 venv/
--- a/.gitignore
+++ b/.gitignore
@@ -129,7 +129,10 @@ celerybeat.pid
 
 # Environments. The suffixed variants hold the same session cookie and proxy
 # credentials as the plain file, and the template is the one that belongs here.
-.env*
+# `.env.*` rather than `.env*`, which also swallowed `.envrc` and left a
+# contributor adding one to wonder why it never showed up in `git status`.
+.env
+.env.*
 !.env.example
 .venv
 env/
--- a/.gitignore
+++ b/.gitignore
@@ -127,8 +127,10 @@ celerybeat.pid
 # SageMath parsed files
 *.sage.py
 
-# Environments
-.env
+# Environments. The suffixed variants hold the same session cookie and proxy
+# credentials as the plain file, and the template is the one that belongs here.
+.env*
+!.env.example
 .venv
 env/
 venv/
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,16 @@ WORKDIR /app
 COPY pyproject.toml uv.lock README.md ./
 RUN uv sync --frozen --no-install-project --no-dev --no-editable --compile-bytecode
 
+# The project itself is installed separately, because installing it runs
+# `setuptools.build_meta` and `uv sync` has no way to constrain that build. The
+# distributions published from CI pin the backend against this same hashed file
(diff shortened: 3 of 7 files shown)


stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

stickerdaniel:
@greptileai review

Issue #483, opened 2026-06-02
[BUG] send_message can create separate DM instead of replying to existing InMail thread


dpearson2699:
## Bug Description

`send_message` can create a separate LinkedIn DM thread instead of replying to an existing InMail/recruiter thread with the same participant.

This is especially risky because the tool returns `sent: true`, and a later scrape of the original thread page may include the new DM's sidebar preview. A caller can incorrectly treat that sidebar preview as verification that the original InMail thread was replied to.

No raw names, profile URNs, thread IDs, screenshots, or message body are included here because the repro came from a real private recruiter conversation.

## Environment

- Package: `linkedin-scraper-mcp`
- Version observed: `4.13.2`
- Runtime: `macos-arm64-host`
- Client: Hermes native MCP client
- Browser/session: logged-in persistent profile

## Steps to Reproduce

1. Start with an existing LinkedIn InMail/recruiter conversation.
2. Read the existing conversation by thread ID:

```json
get_conversation({
  "thread_id": "[existing-inmail-thread-id]"
})
```

3. Send a reply to the same participant using profile-based `send_message`:

```json
send_message({
  "linkedin_username": "[redacted-profile-id-or-slug]",
  "profile_urn": "[redacted-profile-urn]",
  "message": "[approved multi-paragraph reply]",
  "confirm_send": true
})
```

4. Tool returns success:

```json
{
  "url": "https://www.linkedin.com/messaging/compose/?profileUrn=urn%3Ali%3Afsd_profile%3A[redacted]&recipient=[redacted]&screenContext=NON_SELF_PROFILE_VIEW&interop=msgOverlay",
  "status": "sent",
  "message": "Message sent.",
  "recipient_selected": true,
  "sent": true
}
```

5. Open LinkedIn Messaging UI.

## Expected Behavior

For workflows that start from an existing messaging/InMail thread, the send operation should target that existing thread, or the tool should make it impossible to accidentally create a different thread.

Safer options:

- Add a `thread_id` parameter to `send_message`, or add a separate `send_conversation_message` / `reply_to_conversation` tool.
- When `linkedin_username` / `profile_urn` is used and LinkedIn has an existing InMail thread or multiple threads for the participant, return an explicit `thread_id_required` / `ambiguous_conversation` style error instead of composing a profile-based DM.
- Return the actual conversation/thread ID that received the message.
[...]

PR #729, opened 2026-08-15
feat: add reply_to_thread tool for LinkedIn messaging threads

elanthingal:
## Summary
Adds a new `reply_to_thread` tool to reply to an existing LinkedIn messaging thread.

## Changes
- `linkedin_mcp_server/scraping/extractor.py`: adds `reply_to_thread` method
- `linkedin_mcp_server/tools/messaging.py`: registers `reply_to_thread` tool
- `tests/test_tools.py`: adds unit test coverage

## Test Plan
- Existing tests pass
- New `reply_to_thread` test passes

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -103,8 +103,12 @@ celerybeat.pid
 # SageMath parsed files
 *.sage.py
 
-# Environments
-.env
+# Environments. The pattern has to cover the suffixed variants and every
+# directory, because `COPY . .` keeps whatever the context carries in the
+# builder layer, and a `.env.local` holds the same session cookie and proxy
+# credentials as the plain file. That layer is not published, the runtime stage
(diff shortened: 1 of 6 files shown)


elanthingal:
Linking related issue #483: this PR adds  specifically to avoid the profile-based  ambiguity described there, where a new DM can be created instead of replying to an existing InMail/recruiter thread. The new tool targets an existing thread ID directly and now returns the actual outgoing thread reference instead of only checking for visible message text.

elanthingal:
Linking related issue #483: this PR adds `reply_to_thread` specifically to avoid the profile-based `send_message` ambiguity described there, where a new DM can be created instead of replying to an existing InMail/recruiter thread. The new tool targets an existing thread ID directly and now returns the actual outgoing thread reference instead of only checking for visible message text.

PR #738, opened 2026-08-17
feat(profile): return structured top-card location

iller75:
## Summary
- Returns a conservative, structured `location` field when `get_person_profile` can identify exactly one location in the profile top card.
- Ignores ambiguous, unmarked, experience, sidebar, and non-location values rather than guessing.
- Adds DOM and scraper propagation coverage.

## Verification
- The draft is published from clean commit `9788cd93fdf782c887253bc67e53cd439f532acd`.
- Schema metadata verification passed against the locally running streamable-HTTP MCP server: `initialize` accepted protocol `2025-03-26`; one `tools/list` response contained `get_person_profile`.
- Project test suite was not rerun in this invocation.

## Synthetic prompt

> Add a conservative structured top-card location field to `get_person_profile`. Extract only one unambiguous element marked as a LinkedIn location within the section containing the profile h1; omit it on ambiguity or failure. Propagate it through the scraper, document the optional result field, and cover valid, ambiguous, unmarked, sidebar, experience, and non-location cases.

Generated with GPT-5.6 Terra

### DIFF
--- a/linkedin_mcp_server/drivers/browser.py
+++ b/linkedin_mcp_server/drivers/browser.py
@@ -18,6 +18,7 @@ from linkedin_mcp_server.common_utils import harden_linkedin_tree, secure_mkdir
 from linkedin_mcp_server.core import (
     AuthenticationError,
     BrowserManager,
+    NetworkError,
     detect_auth_barrier_quick,
     detect_rate_limit,
     goto_reporting_proxy_errors,
@@ -252,15 +253,13 @@ async def _feed_auth_succeeds(
         # disk and the log is what users paste into issue reports.
         await _log_feed_failure_context(browser, detail)
         if barrier is None:
-            # Nothing loaded and no barrier, so nothing proves the session is
-            # dead -- and with a proxy in front, the most likely cause is the
-            # proxy. Wrong credentials in particular produce no proxy error code
-            # at all: Chromium retries the 407 challenge until the navigation
-            # times out (verified against a local authenticating relay), so the
-            # marker check above cannot catch it. Reporting False would hand the
-            # caller an AuthenticationError, whose recovery moves the stored
-            # profile aside and starts a login through the same broken proxy.
+            # A navigation exception without a login/checkpoint/authwall marker
+            # proves only that the feed transport failed. It must remain a
+            # retryable network failure: converting it to ``False`` makes the
+            # caller manufacture AuthenticationError and quarantine live source
+            # cookies without authentication evidence.
             raise_if_proxy_configured(exc)
+            raise NetworkError("Feed navigation failed; try again.") from exc
         return False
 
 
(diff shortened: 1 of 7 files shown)


iller75:
Added `bb3c346fa3c666513492a4c22980bbfe2b44d257`: an HTTP 400 (or other feed navigation failure) without a detected login/checkpoint/authwall barrier now raises retryable `NetworkError` rather than returning an auth failure that quarantines saved source state.

Coverage: `tests/test_browser_driver.py` (41 passed), `tests/test_dependencies.py` (29 passed), combined 70 passed; `ruff format --check`, `ruff check`, `ty check`, and `pre-commit --all-files` passed.

`uv run pytest -q`: 1947 passed, 39 skipped, 1 failed. The sole failure is pre-existing and was reproduced unchanged on clean parent `9788cd93fdf782c887253bc67e53cd439f532acd`: `tests/test_daemon_descriptor.py::TestEndpointSpelling::test_an_endpoint_that_composes_into_nothing_usable_is_refused[127.0.0.1.-/mcp]` (expected `DescriptorError`, did not raise).

iller75:
## Post-restart session probe fix (commit f3c9a16)

**Problem:** After service restart, `_authenticate_existing_profile` called `_feed_auth_succeeds`, which navigated to `/feed/`. On fresh Chromium in some environments LinkedIn returns a generic HTTP 400. `bb3c346` correctly classified this as retryable `NetworkError`, but the caller had no retry — so the first tool call after every restart failed.

**Fix (f3c9a16):** Two new functions added to `linkedin_mcp_server/drivers/browser.py`:

- `_probe_auth_via_profile_nav(browser)` — navigates to `https://www.linkedin.com/in/me/`, checks for auth barrier only. Same route proven by `get_my_profile`; does not receive the `/feed/` HTTP 400.
- `_probe_post_restart_session(browser)` — wraps the above with bounded retry: max 3 attempts, 2s delay. Returns True on first success. Returns False immediately on auth barrier (definitive signal, no retry). Raises `NetworkError` only after all 3 attempts fail.

`_authenticate_existing_profile` now calls `_probe_post_restart_session` instead of `_feed_auth_succeeds`. `_feed_auth_succeeds` is unchanged and still used by bridge/import paths.

**Tests:** New file `tests/test_post_restart_session_probe.py` (6 tests):
- RED: confirms `_feed_auth_succeeds` had no retry (documents old behavior)
- GREEN: retry succeeds on attempt 2; auth barrier stops retry immediately; 3 failures raises NetworkError; URL is `/in/me/` not `/feed/`; startup probe does not call `_feed_auth_succeeds`

Two existing tests in `test_browser_driver.py` updated to reflect new behavior (probe navigates once to `/in/me/`, no remember-me at startup; HTTP 400 error message updated to match new probe path).

**Full suite:** 1936 passed, 21 skipped, 1 known baseline failure (`test_daemon_descriptor::test_an_endpoint_that_composes_into_nothing_usable_is_refused[127.0.0.1.-/mcp]` — identical to pre-fix baseline). No regressions introduced.

Issue #740, opened 2026-08-18
--login waits out its timeout when sign-in completes in a different tab (valid li_at already written to profile)


dannybloomfield:
### Summary

`wait_for_manual_login()` polls a single `Page` reference. If the user completes sign-in in a **different tab** than the one the server opened, the tracked page never changes, `is_logged_in()` keeps returning `False`, and `--login` waits out its full timeout with no indication that anything is wrong — while a perfectly valid session (including `li_at`) has already been written into the profile the server itself created.

The detector is behaving correctly given what it's watching. The gap is that nothing notices the session exists, and nothing tells the user which window matters.

### Environment

| | |
|---|---|
| OS | macOS, Darwin 24.6.0, x86_64 |
| Version | `mcp-server-linkedin` 4.22.0 (also read from `:latest` image) |
| Install | `uvx mcp-server-linkedin@latest` |

### What happened

`uvx mcp-server-linkedin@latest --login --login-timeout 0`, then signed in. The log never advanced past:

```
INFO core.auth: Please complete the login process manually in the browser. Waiting with no time limit (LOGIN_TIMEOUT=0)...
```

Observed for >10 minutes. No `cookies.json` / `source-state.json` written; `SIGTERM` did not flush a session; `--status --no-auto-import` afterwards reported `❌ No valid source session found`.

But the login had genuinely succeeded. Querying the server's own profile store:

```console
$ sqlite3 ~/.linkedin-mcp/profile/Default/Cookies \
    "select count(*) from cookies where host_key like '%linkedin%';"
17

$ sqlite3 ~/.linkedin-mcp/profile/Default/Cookies \
    "select name, datetime(last_update_utc/1000000-11644473600,'unixepoch','localtime') from cookies where name='li_at';"
li_at|2026-08-17 14:29:21
```

Browser launched 14:25:59; `li_at` written 14:29:22. So a valid session sat in the profile for the remainder of the wait, unnoticed.

### Why I think it's the tab

During that window, Chromium renderer processes for the managed browser went from 9 to 14 — consistent with additional tabs being opened during sign-in. Reading the source:

```python
async def wait_for_manual_login(page: Page, timeout: int = 300000) -> None:
    while True:
        if await resolve_remember_me_prompt(page):
            ...
        if await is_logged_in(page):
            logger.info("Manual login completed successfully")
            return
        await asyncio.sleep(1)
```

[...]

PR #741, opened 2026-08-18
docs: note that --login must be completed in the window it opens

dannybloomfield:
Docs-only follow-up to #740, opened first per CONTRIBUTING.

`wait_for_manual_login()` polls the single `Page` reference the server opened. If a user completes sign-in in a different tab or window, `is_logged_in(page)` never becomes true and `--login` waits out its whole timeout, even though a valid `li_at` has already been written to the profile. Nothing in the README currently says which window matters, so the failure looks like a broken login rather than a misdirected one.

This adds:

- a clarifying clause on the `--login` flag description (both places it is documented)
- a troubleshooting bullet under each install section's **Login issues** list, matching the existing convention of repeating those bullets per section

No behaviour change. 6 additions, 2 deletions.

Happy to trim this to a single section, or reword, if you would rather keep the duplication down.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ When you set up or maintain this server, verify its entry in the MCP client conf
 
 **CLI Options:**
 
-- `--login` - Open browser to log in and save persistent profile
+- `--login` - Open browser to log in and save persistent profile. Complete the sign-in in the window that opens; the wait watches that page.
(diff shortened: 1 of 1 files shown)


stickerdaniel:
Thanks! Closing: `--login` now checks the auth cookie across the whole browser context, so any tab works (#740).

PR #727, opened 2026-08-13
fix(jobs): Read redesigned search results

truehhart:
Fixes #725. Supersedes #726 (same change; branch was renamed to follow the repo's naming convention, which auto-closed the old PR).

## Problem

LinkedIn now 302-redirects `/jobs/search/` to `/jobs/search-results/` (their new AI search experience). This breaks `search_jobs` in three ways:

1. The URL guard only accepted `/jobs/search/`, so it skipped job ID extraction after every redirect.
2. The new result cards store job IDs in a `componentkey` attribute (`job-card-component-ref-<id>`), not in `/jobs/view/` anchor hrefs. The old selector found nothing.
3. The redirect drops the `location` query parameter. LinkedIn falls back to the account's default location.

## Changes

- **URL guard**: accepts both `/jobs/search` and `/jobs/search-results` paths. Also checks the host is `www.linkedin.com`.
- **Job ID extraction + sidebar scroll**: matches both classic `/jobs/view/` anchors and the new `componentkey` cards, deduplicated.
- **Location handling**: navigates with plain keywords first. If the redirect to the new route is detected and a location was requested, re-fetches once with the location folded into the keywords (e.g. `"python in Berlin"`), which the AI search parses correctly. Classic route behavior is unchanged. Cost is one extra navigation on page 1, only when redirected. Happy to take a different approach here if you prefer.
- DOM dependencies documented in docstrings per CONTRIBUTING.

## Verification

- **Unit tests**: redirected route, location re-fetch, classic route keeps plain keywords, host guard.
- **Browser-DOM tests**: both card formats, dedup, scroll assertion.
- **Live check via MCP stdio**: Berlin search returned 99+ results and 25 job IDs with location respected.
- **Full suite**: 1975 passed, 11 skipped. Pre-commit green (ruff, ruff-format, ty).

PR #743, opened 2026-08-18
fix(jobs): Load a full page of job cards

### LINKED ISSUE #742: [BUG] search_jobs returns 11 of 25 jobs per page
### Setup

`uv run -m linkedin_mcp_server --transport streamable-http --log-level DEBUG` on macOS, authenticated profile at `~/.linkedin-mcp/profile/`.

### What Happened

`search_jobs` returns roughly 11 jobs per page where LinkedIn has 25.

For `AI Engineer` in Noida, LinkedIn reports 341 results across 14 pages, so 25 per page, and `_PAGE_SIZE = 25` (`scraping/extractor.py:98`) matches that. A two-page call returned 22 job IDs in total. The page text carries exactly 22 job cards, so extraction is complete against what rendered. The cards themselves never loaded.

The sidebar scroll is where they are lost. `scroll_job_sidebar` (`core/utils.py:89`) stops as soon as `container.scrollHeight` is unchanged after a single wait, and the call site passes `pause_time=0.5` (`scraping/extractor.py:3138`). Lazy loading regularly needs longer than half a second, and one late batch ends the loop for good. The log agrees: two successful scrolls per page, then the loop exits well below `max_scrolls=5`. A confirmation round before giving up would cover the slow batch that currently ends it.

Three runs, same account, same query:

| date | max_pages | job_ids | cards in text | page header |
| --- | --- | --- | --- | --- |
| 14.08. | 1 | 10 | 10 | 300 results |
| 15.08. | 1 | 9 | 9 | 316 results |
| 16.08. | 2 | 22 | 22 | 341 / 347 results |

Second finding from the same run, possibly separate: `job_ids` counted 22 while `references.search_results` held 14 job entries. On the single-page runs the two matched exactly (10/10 and 9/9), so references appear to be lost when pages are combined.

This is independent of #725. It reproduces on the classic job search route, and the account never saw the redesigned one.

### Steps to Reproduce

1. Call `search_jobs(keywords="AI Engineer", location="Noida", max_pages=2)`.
2. Count `job_ids`: 22, against the 14 pages of 25 that the page itself reports.

### Logs

```shell
07:46:04 - scraping.extractor - DEBUG - _navigate_to_page: target=https://www.linkedin.com/jobs/search/?keywords=AI+Engineer&location=Noida
07:46:08 - core.utils - DEBUG - Scrolled job sidebar 2 times
07:46:08 - scraping.extractor - DEBUG - LinkedIn reports 14 total pages
07:46:10 - scraping.extractor - DEBUG - _navigate_to_page: target=https://www.linkedin.com/jobs/search/?keywords=AI+Engineer&location=Noida&start=25
07:46:14 - core.utils - DEBUG - Scrolled job sidebar 2 times
```

stickerdaniel:
Closes #742

The job sidebar scroll gave up as soon as the container had not grown after a single fixed 0.5s sleep, so one batch arriving later than that ended the page permanently. Measured against live LinkedIn: a search that the page itself reports as 14 pages of 25 returned 11 cards per page, twice in a row, with `Scrolled job sidebar 2 times` in the log both times. Pagination was never the problem, `_PAGE_SIZE = 25` matches what LinkedIn does; the cards simply never loaded.

How long a batch takes belongs to the user's connection, so no constant is right: raise it and every fast client pays the worst case on every page, keep it low and slow clients lose more than half the results. Each round now polls for the next batch, and a round that sees nothing scrolls once more at the full `settle_timeout` before the list counts as exhausted. That confirmation round is what makes a single slow batch survivable, including the very first one, where no measurement exists yet. Later rounds start from three times what the previous batch actually took, floored at `min_budget`, so a fast connection stays fast: measured 0.17-0.35s for a full page of 25.

A wall-clock `deadline` bounds the whole call. `max_pages` reaches 10 and `tool_timeout_seconds` defaults to 180, so a page that keeps trickling could otherwise have spent the tool's entire budget. Counting and container discovery are both scoped to the rail: the detail pane carries its own `/jobs/view/` permalink, and walking up from the first match in the document found no scrollable ancestor at all when that link came first, which disabled the scroll entirely.

The DOM tests drive a synthetic sidebar whose per-batch latency is a list, so a fixture can vary it the way a connection does. Covered: an 800ms batch (the old 0.5s budget stops at 5 of 25), a spike after two fast batches (15 of 25 without the confirmation round), a slow first batch, a stray link outside the rail, the deadline, and an exhausted list. Full suite 1983 passed, ruff and ty clean.

Sequencing with #727: it makes the card set a union of `/jobs/view/` anchors and the redesigned `componentkey` cards, while this PR builds counting, the wait-for-cards guard and the container walk on the single anchor selector. Whichever lands second has to move all three to the union selector, otherwise the redesigned route never scrolls at all. Not a textual rebase.

## Synthetic prompt

> `search_jobs` returns about 11 jobs per page although LinkedIn paginates at 25. Find out why, fix it in `scroll_job_sidebar`, and make the fix hold for both fast and slow connections rather than tuning a constant. Add DOM tests that fail against the old behaviour, and review the result adversarially for what a network with uneven latency does to it.

Generated with Claude Opus 5 (high)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -148,18 +148,19 @@ Optional additional keys:
 
 ## Tests
 
-- **A test that cannot fail is not a test.** Run a new test against the
-  broken state it is meant to catch before committing it. One that passes
-  there asserts something every implementation satisfies (a count that holds
-  either way, a branch merely touched) and has to be rewritten to assert the
-  behaviour, usually against the log line or the elapsed time. Two of the ten
-  tests in `tests/test_job_sidebar_scroll_dom.py` shipped that way and only
-  review caught them.
+- **A test that cannot fail is not a test.** Before committing one, mutate
+  the code it covers and watch it fail. A test that survives the mutation
+  asserts something every implementation satisfies (a count that holds
+  either way, a branch merely touched), and the usual repair is to assert
+  the log line, the elapsed time, or a count scoped to the thing under
+  test. Fixtures that reload on every scroll event mask a premature stop:
+  slow them down until one batch lands per round, or the mutation survives
+  for the wrong reason.
 - **Browser-DOM tests belong where the unit suite mocks `page.evaluate`.**
   Extractor JS never executes under a mock, so a `browser_dom` test is its
-  only coverage. Elsewhere prefer a unit test: a browser fixture is a claim
-  about LinkedIn's own behaviour, and a wrong claim costs more than the
-  missing test would have.
+  only coverage. Prefer a unit test elsewhere, and keep in mind that a
+  fixture imitating LinkedIn's markup is a claim about LinkedIn, while one
+  driving a synthetic container is a claim about the algorithm only.
 
 ## Verifying Bug Reports
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -146,6 +146,21 @@ Optional additional keys:
 - `job_ids: [id, ...]` (search_jobs and get_saved_jobs)
 - `references["feed"]` (get_feed only) — every entry is `kind: "feed_post"`; non-post anchors (sidebar profiles, employer logos) are filtered. URLs may carry either `/feed/update/<urn>/` (DOM-anchor-derived) or `/posts/<slug>` (SDUI-derived) form; both are valid LinkedIn permalinks. Cap is 50 entries, matching `get_feed`'s `num_posts` ceiling.
 
+## Tests
+
+- **A test that cannot fail is not a test.** Run a new test against the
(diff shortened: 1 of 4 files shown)


stickerdaniel:
Superseded by #751.

The premise here was wrong, and four review rounds checked the implementation without ever checking it. The page renders 11 jobs, not 25, so there was never a page of cards to recover by scrolling harder; measurements are in #742. #751 pages by what each navigation actually rendered and keeps the one part of this PR that measures a real gain, the confirmation round that stops the scroll from giving up after a single 0.5s look.

Issue #809, opened 2026-08-27
[BUG] PGID reuse can misdirect browser cleanup


stickerdaniel:
**Setup:** POSIX process-tree cleanup under rapid PID and process-group churn.

## What Happened

Browser cleanup validates a recorded process-group leader or member identity and then signals the numeric PGID in a separate system call. The crash guardian retains only numeric PGIDs.

If the marked group disappears after validation and the kernel reuses its leader PID as a new PGID before `killpg`, cleanup can signal an unrelated process group owned by the same account. The current tests cover reuse before validation, not the check-to-signal interval.

The signal path needs an identity-stable mechanism where the platform provides one, or a design that keeps a group member unreaped until signaling is complete.

## Steps to Reproduce

1. Register a synthetic detached process group and record its start identity.
2. Let that group exit immediately after the cleanup identity check.
3. Reuse the numeric leader PID for an unrelated process group before signaling.
4. Observe that `killpg` targets the replacement group.

Amiirhosseini:
Hi @stickerdaniel — would you like me to work on this?

**Idea / plan:** Fix PGID reuse causing browser cleanup to target the wrong process group. Use a fresh process group per browser session or validate PGID before cleanup signals. Add regression test for PGID collision scenario.

Happy to open a PR once you confirm I should take this.

stickerdaniel:
Audit status: blocked, keeping this open. Current code still validates process identity and later signals the numeric process group in separate operations. Existing tests cover reuse visible before validation. A conclusive run must force the kernel to recycle the exact leader PID into a new PGID inside that interval, which I could not reproduce safely. Thanks for offering to work on it.

Issue #808, opened 2026-08-27
[BUG] Windows owner crash can outrun Job drain


stickerdaniel:
**Setup:** shared daemon mode on Windows while Chromium is active.

## What Happened

The owner’s final named-Job handle and profile-lock handle both close during native process teardown. `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` starts terminating Job members, but process termination is asynchronous and no surviving process waits for `ActiveProcesses == 0`.

A replacement owner can therefore acquire the profile lease while Chromium from the crashed owner is still terminating. The controlled hard-exit path drains the Job before releasing locks; native crashes have no equivalent fence.

Windows owner containment needs an external crash guardian or another coordination lock that remains held until the adopted Job reports no active processes.

## Steps to Reproduce

1. Start an owner in its named Job with a persistent synthetic descendant.
2. Hard-terminate the owner process.
3. Immediately contend for the same profile lease from another process.
4. Observe whether the lease becomes available before the descendant has exited and the Job is empty.

Amiirhosseini:
Hi @stickerdaniel — would you like me to work on this?

**Idea / plan:** Reproduce Windows owner crash outrunning Job drain. Ensure the owner process waits for child Job completion (or handles `WAIT_OBJECT_0`/`STILL_ACTIVE` correctly) before tearing down shared state. Add tests for crash-during-drain scenario.

Windows user, already forked the repo. Please let me know if I can be assigned.

stickerdaniel:
Audit status: blocked, keeping this open. Exact-commit Windows CI covers controlled hard exit and Job draining, but does not hard-terminate the owner and immediately contend for the profile while Job members are still terminating. [PR #825](https://github.com/stickerdaniel/linkedin-mcp-server/pull/825) intentionally leaves this native Windows crash case here. Thanks for offering to work on it.

Issue #807, opened 2026-08-27
[BUG] Direct browser crashes release profiles early


stickerdaniel:
**Setup:** manual `--login` or browser-cookie import on any supported OS.

## What Happened

These paths acquire the profile lease and create `BrowserManager` directly. They bypass the daemon owner’s POSIX crash guardian and Windows Job containment.

A native crash or external hard kill can therefore release the kernel profile lease while detached Chromium remains alive. Another login or import can acquire the lease and rotate or open the same live profile. Concurrent Chromium instances can overwrite profile stores and lose the authenticated session without a clear error.

The direct browser paths need their own crash-containment boundary that keeps replacement processes off the profile until the browser tree is proven gone.

## Steps to Reproduce

1. Start a direct browser path with a synthetic persistent-browser child.
2. Hard-kill the process after the child detaches.
3. Acquire the same profile lease from another process.
4. Observe that the lease succeeds while the detached child remains alive.

Amiirhosseini:
Hi @stickerdaniel — would you like me to work on this?

**Idea / plan:** Reproduce the race where a direct browser crash releases profiles before the owner Job drain completes. Add synchronization so profile cleanup waits for the Job object drain on Windows, with regression tests simulating early browser exit.

I'm on Windows and already contributing to this repo (fork + #675). Happy to take this if assigned.

stickerdaniel:
Re-ran the direct POSIX crash path with a real profile lease, crash guardian, detached marked browser, and `SIGKILL`. The contender reacquired the lease only after the browser was dead. [PR #825](https://github.com/stickerdaniel/linkedin-mcp-server/pull/825) integrated that guardian. The distinct native Windows crash-drain case remains tracked in [issue #808](https://github.com/stickerdaniel/linkedin-mcp-server/issues/808). Thanks for offering to work on this.

PR #825, opened 2026-08-30
fix(daemon): Contain managed process trees

### LINKED ISSUE #789: [BUG] Startup rollback can orphan browser installers
**Setup:** managed browser runtime on POSIX. The direct process behavior was measured on macOS; the Patchright-specific behavior is recorded in `bootstrap._stop_installer`.

## What Happened

The daemon starts background browser setup before publishing its ready verdict. That setup launches `python -m patchright`, which launches Node descendants for the download.

Startup rollback currently kills only the direct owner PID. The installer cleanup has the same boundary: `_stop_installer` kills the Python wrapper while its Node descendants continue downloading and refreshing the shared browser-cache lock. A later startup can queue behind an installer whose owner no longer exists.

This surfaced while reviewing #781. That fix now avoids adding new kill paths after configuration handover, but the existing silent-startup timeout can still reach this behavior.

A cross-platform repair needs to contain the installer tree or move managed setup outside the rollback window. The existing source comment records that putting the installer in its own session caused a Windows crash and broke terminal signal behavior, so this needs its own design.

## Steps to Reproduce

1. Start a detached child that starts a sleeping grandchild.
2. Pass the direct child to `daemon_election._stop_child`.
3. Check both PIDs after `_stop_child` returns.

Measured result:

```shell
owner_returncode=-9
grandchild_alive_after_owner_stop=True
```

The application path is the managed-browser startup path through `browser_lifespan`, `start_background_browser_setup_if_needed`, and `_run_patchright_install`.

See also: #781

stickerdaniel:
Closes #789

Stopping failed owners or installs previously killed only Python. Patchright descendants could retain locks or outlive profile ownership.

Managed installs and owners use OS containment. POSIX tracks browser groups, drains zombies, blocks supervisor startup hooks and holds profile leases until live members exit. Hard exits free the election before the unbounded profile-safe drain. Windows gives each launch a kill-on-close Job within outer Jobs. Cancellation keeps shutdown verdicts; reused managers mint fresh attribution. Two authenticated frames are written without newline translation, which Windows would otherwise turn into a refusal, and the drains spare the gate that relays an exit status. The installer's target no longer inherits the worker's control pipe: on Windows it queued behind the lease read and never ran a line, so a managed install never returned. The CLI login and the session import start and release a crash guardian of their own, because each builds its browser outside the shared launch and one guardian exists per process. A stop after Job assignment falls back to kill-on-close, so a Job left open by a failing termination cannot outlive the election it ended. A start that fails now reports the close it proved, which only the exit used to record and which the login reads before deciding the profile's fate; a login that ends before the browser opens no longer reads that decision off a manager it never handed anything to. Direct POSIX crashes, Windows crash drainage and PGID reuse remain #807, #808 and #809.

Verified with the full suite (2,569 passed, 33 skipped), Ruff, ty, pre-commit, browser tests, Linux mutations and real process trees on Windows, macOS and Linux arm64.

## Synthetic prompt

> Contain daemon owners, managed browser installers and detached browser processes on POSIX and Windows so cancellation and owner death cannot leave descendants behind or release profile ownership early.


stickerdaniel:
* **#827** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/827?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#826** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/826?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#825** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/825?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/825?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
### Merge activity

* **Aug 31, 2:12 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/825).
* **Aug 31, 2:13 PM UTC**: @stickerdaniel merged this pull request with [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/825).

Issue #823, opened 2026-08-30
[BUG] Election retries a bind that cannot succeed


stickerdaniel:
## Setup

Daemon sharing enabled, on a host where this frontend cannot create a loopback listener at all: a sandbox that forbids one, an exhausted descriptor limit, or a local endpoint-security policy that refuses the socket.

## What Happened

`_spawn` binds its control rendezvous first, before any child exists, because the child needs that address in its configuration record. When `ControlListener.open()` raises, `obtain_owner()` classifies the attempt as `_Attempt.FAILED` and applies the failed-child retry policy: it keeps spawning until the 90-second election budget is spent.

That state means something else. A failed child may have released the daemon lock, so another candidate can win it and observing further is worth the wait. A parent-side bind that cannot succeed has no child and no lock, and repeating it cannot change the answer. Because `_obtain_shared_owner()` runs during server construction, direct-browser fallback waits out the full budget, and a host or supervisor with a shorter startup timeout kills the server before it gets there.

What the retry does still buy is descriptor observation: a sibling frontend on the same account may publish an owner this one can attach to, and that is worth keeping. So the fix is not simply to fail fast. It needs a state that stops starting children while observation continues, and a decision about whether a frontend that can never publish should spend the whole budget waiting for a sibling that may not exist. Both are choices rather than repairs, which is why this is filed rather than fixed.

## Steps to Reproduce

1. Enable daemon sharing.
2. Make loopback listener creation fail for the process, for example by exhausting the descriptor limit.
3. Start the server.
4. Observe repeated start attempts and roughly 90 seconds before direct-browser fallback.

stickerdaniel:
Re-ran the reproduction with `ControlListener.open` raising `OSError`. Current main attempted listener creation once and returned direct-browser fallback in 0.004s instead of spending the election budget. [PR #827](https://github.com/stickerdaniel/linkedin-mcp-server/pull/827) added this bounded pre-spawn failure path. Closing as resolved.

Issue #821, opened 2026-08-30
[BUG] Windows account home must be DACL-protected to run a daemon


stickerdaniel:
**Setup:** shared daemon mode on Windows with a redirected or network account home, such as `\\fileserver\profiles\alice`.

## What Happened

`prepare_daemon_state` asks `verify_children_cannot_be_replaced(home)` before it creates anything under the account home, and that check requires the home's DACL to be protected. A local `C:\Users\alice` is, because Windows creates a profile root with an explicit non-inherited ACL. A redirected or UNC profile commonly is not: it inherits a user-only DACL from an administrator-controlled parent, which is safe but not protected.

The refusal lands before `.mcp-server-linkedin-v2` exists, so there is nothing to delete and no alternative state root to point at. The server still runs, falling back to an in-process browser for every call, but shared daemon mode is unavailable for that account until someone changes the ACL inheritance on the profile directory.

The check has a `require_protected=False` mode, used for the installer temporary parent, that accepts an inherited DACL and asks `verify_ancestry_cannot_be_replaced` about the whole chain instead. Whether that is the right shape for a token-bearing state root, and whether the ancestry walk answers correctly for a UNC path, needs measuring on a real domain-joined machine.

## Steps to Reproduce

1. Use a Windows account whose profile is redirected to a file share.
2. Confirm the profile directory's DACL is inherited rather than protected.
3. Start the server with `--daemon`.
4. Observe the refusal naming the profile directory, and every election falling back.

stickerdaniel:
Audit status: blocked, keeping this open. The strict ancestry check still rejects unprotected inherited DACLs, while Windows CI covers only simulated ACLs. The issue needs a domain-joined account with a redirected UNC home and inherited administrator-controlled ancestry for a real verdict.

Issue #819, opened 2026-08-29
[BUG] Predecessor owner started by a current frontend loses its log


stickerdaniel:
**Setup:** shared daemon mode with a frontend from the current release still loaded while the installed package is rolled back to the predecessor release.

## What Happened

The predecessor owner expects its frontend to have opened `daemon.log` as `stderr`, and redirects its own `stdout` onto that descriptor. The current frontend instead spawns with `stderr=subprocess.PIPE`, because opening the log moved into the owner process, where a blocking state store cannot pin a frontend that has to stay killable.

A predecessor owner started this way therefore writes its whole lifetime into a pipe. The bootstrap reader drains and discards every non-bootstrap line, and once the frontend exits the owner writes into a pipe with no reader at all. Its published `log_path` names a file this startup never created, so `--daemon-logs` and every later diagnosis find nothing or find a stale file from an earlier generation.

The owner stays attachable and keeps serving, so this is lost observability rather than lost service. Repairing it needs a way to give a predecessor owner a real log without moving log creation back into the frontend. `tests/predecessor_owner.py` does not model the stream redirection, so the compatibility test stays green through this.

## Steps to Reproduce

1. Start a frontend from the current release and keep it loaded.
2. Replace the installed package with the predecessor release.
3. Have the frontend spawn an owner, which now runs the predecessor protocol.
4. Let the owner publish and outlive the frontend.
5. Observe that `daemon.log` is missing or stale while the owner is serving.

Issue #818, opened 2026-08-29
[BUG] Predecessor owner can be stopped after it published


stickerdaniel:
**Setup:** shared daemon mode with a frontend from the current release still loaded while the installed package is rolled back to the predecessor release.

## What Happened

The current frontend starts an owner from the installed package, which is now the predecessor. That protocol has no prepare and commit boundary: the owner publishes its descriptor and token on its own authority and only afterwards writes `READY`. The frontend recognises the published verdict and stops claiming termination authority from that point, but it has no signal for the interval before it.

If the child is suspended, starved or interrupted between publication and `READY`, `_await_prepared` exhausts its budget and the frontend calls `stop_child()`. The canonical descriptor then names an endpoint the frontend itself killed. A later election probes that endpoint, finds it dead and replaces the descriptor, so the daemon recovers, but this startup loses the guarantee the current protocol provides.

Closing this needs a way for the frontend to recognise a predecessor publication that carries no commit record, for example by matching the canonical descriptor against the child it started before deciding to stop it. `tests/predecessor_owner.py` writes `READY` without publishing anything, so it cannot exercise the interval either.

## Steps to Reproduce

1. Start a frontend from the current release and keep it loaded.
2. Replace the installed package with the predecessor release.
3. Have the frontend spawn an owner, which now runs the predecessor protocol.
4. Suspend the child after it publishes its descriptor and before it writes `READY`.
5. Observe the startup budget expire and the frontend stop an owner whose descriptor is already canonical.

PR #827, opened 2026-08-30
fix(daemon): Commit owner publication atomically

### LINKED ISSUE #790: [BUG] Daemon startup lacks an atomic commit boundary
**Setup:** daemon mode on any platform. The lock wedge is POSIX-specific because the frontend transfers the held descriptor; the publication race applies wherever another frontend can read the descriptor.

## What Happened

Daemon startup has no atomic commit boundary between the private parent-child handshake and global descriptor publication.

The child publishes its descriptor before writing `ready`. If the starting frontend loses the verdict through `KeyboardInterrupt`, an internal reader error, or the startup timeout, it cannot tell which state the child reached:

- Killing the child can terminate a published owner that another frontend has already attached to.
- Leaving the child can preserve a pre-publication process that holds the inherited lock and will never serve.

Both outcomes are reachable. The existing timeout chooses the first; interrupted waits choose the second. A descriptor lookup before acting cannot close the race because the child can publish between the lookup and the kill.

The startup protocol needs a two-phase commit or child-side lease. Global publication must happen only after the parent commits the child lifetime, and a child whose parent disappears before that commit must terminate without publishing. The successful owner must still outlive the frontend afterwards.

## Steps to Reproduce

1. Start an owner child with the inherited lock descriptor.
2. Suspend it before descriptor publication, then interrupt the frontend while it waits for `ready`.
3. The detached child survives with the lock and no published endpoint.
4. In a second run, let the child publish, suspend it before the `ready` write is observed, and attach another frontend through the descriptor.
5. Let the first frontend's existing startup timeout expire.
6. The first frontend kills the owner while the second frontend is using it.

No descriptor probe in the frontend makes both sequences safe without an additional protocol state.

See also: #781, #789

stickerdaniel:
Closes #790

Daemon startup exposed a child before publication was settled. An interruption could kill a published owner or strand its lock.

The child prepares and publishes after validation. Child I/O preserves storage fallback, while ambiguous results retain the lock until publication is proved. A nonce-authenticated loopback channel drains, unblocks aborted reads and paces retries past generations. Windows stages state, pins ancestry, accepts volume-root homes and leaves timed-out pipes with their parked readers. It also closes the lock file before unlinking it, which is the only order Windows allows, and refuses a link naming a directory as the directory it names. The declared Python floor moves to 3.12.4, where Windows first gives a private directory a restricted access list. Predecessor migration, publication and logging remain #810, #818 and #819; unprotected home directories, an unrepeatable bind retry, remote commit pins and diagnostics remain #821, #823, #796, #800 and #801.

Verified with the full suite (3,030 passed, 60 skipped), Ruff, ty and pre-commit. Windows contracts cover 3.12.4 through 3.14.

## Synthetic prompt

> Add an atomic parent-child commit boundary to daemon startup, including mixed-version handover, ambiguous publication, Windows state and concurrent elections.


stickerdaniel:
* **#827** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/827?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/827?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#826** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/826?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#825** <a href="https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/825?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

stickerdaniel:
### Merge activity

* **Aug 31, 2:12 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/827).
* **Aug 31, 2:20 PM UTC**: [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/827) rebased this pull request as part of a merge.
* **Aug 31, 2:25 PM UTC**: @stickerdaniel merged this pull request with [Graphite](https://app.graphite.com/github/pr/stickerdaniel/linkedin-mcp-server/827).

stickerdaniel:
Follow-up: #838 tracks the intermittent Windows CPython 3.12.4 access violation behind the narrowly scoped stress-test skip.

Issue #812, opened 2026-08-28
[BUG] Flaky browser-DOM test: test_loads_full_page_on_a_slow_connection


stickerdaniel:
### Setup

CI, `ubuntu-latest` runner, `uv run pytest -n auto`.

### What Happened

`tests/test_job_sidebar_scroll_dom.py::TestSidebarScroll::test_loads_full_page_on_a_slow_connection` failed with `assert 5 == 25` on run 33146106873, then passed on a rerun of the same commit. `main` was green on the surrounding runs, and the test passes locally, so this is timing, not a regression.

The fixture delays every batch by 800ms while the test calls `scroll_job_sidebar(settle_timeout=0.6)`. That combination is deliberate: 600ms < 800ms, so the first wait has to miss and the confirmation round at the full `settle_timeout` is what buys the batch. But it leaves only `2 * 600 - 800 = 400ms` of configured slack, and the failure is asymmetric. The budget is wall-clock (`Date.now()`), while the batch is a page `setTimeout`; when the runner starves the page's event loop the timer slips late and the budget does not. `assert 5 == 25` means zero successful rounds, so even the very first 1200ms window missed its 800ms timer.

The sibling tests are not exposed the same way. `test_survives_a_slow_first_batch` has `2 * 2000 - 2800 = 1200ms` of configured slack and one slow batch; this test has 400ms across four slow growth batches, so it has both the thinnest margin and the most chances to lose it.

### Steps to Reproduce

Load-dependent, and it does not reproduce on a 24-core machine (12/12 passes under 48 competing `yes` processes). The margin is what is reproducible: the fixture and the scroll loop are deterministic apart from timer scheduling, and the numbers above come straight from `settle_timeout`, the fixture's `delays`, and the two `waitForGrowth` calls in `linkedin_mcp_server/core/utils.py`.

### Logs

```
FAILED tests/test_job_sidebar_scroll_dom.py::TestSidebarScroll::test_loads_full_page_on_a_slow_connection - assert 5 == 25
```

Amiirhosseini:
Hi @stickerdaniel — would you like me to work on this?

**Idea / plan:**
1. Analyze the timing budget in `test_loads_full_page_on_a_slow_connection` (800ms batch delay vs 600ms settle timeout).
[...]

PR #832, opened 2026-08-31
fix(references): Add missing section contexts

### LINKED ISSUE #830: [BUG] References from certifications, skills, projects and employees have no context
### Setup

v4.23.1, `main` at `0cd1e5f`. Not environment-specific — reproduces from a direct call to `build_references`.

### What Happened

`_SECTION_CONTEXTS` in `scraping/link_metadata.py` predates `certifications`, `skills` and `projects` joining `PERSON_SECTIONS`, and `employees` never had an entry. References from those four reach callers with no `context` key while every sibling section names itself.

There is a second effect: `_reference_score` scores a context-less reference strictly lower, so `_choose_better_reference` resolves duplicate-URL collisions against these sections during cross-page dedupe.

Nothing ties the context table to `PERSON_SECTIONS`, which is how three sections reached `main` without an entry.

### Steps to Reproduce

```python
from linkedin_mcp_server.scraping.link_metadata import build_references

raw = [{"href": "https://www.linkedin.com/company/aws/", "text": "Amazon Web Services"}]

for section in ("experience", "certifications", "skills", "projects", "employees"):
    print(section, build_references(raw, section)[0])
```

`experience` carries `'context': 'experience'`. The other four carry no `context` key.

### Logs

n/a — not a runtime failure.

ConnorMoss02:
Closes #830

`_SECTION_CONTEXTS` was written before `certifications`, `skills` and `projects` joined `PERSON_SECTIONS`, and `employees` never had an entry, so references from those four reached callers with no `context` key while every sibling section named itself.

Beyond the missing key, `_reference_score` scores a context-less reference strictly lower, so `_choose_better_reference` resolved duplicate-url collisions against these sections during cross-page dedupe.

The `_REFERENCE_CAPS` entries record the value that already applied through `_DEFAULT_REFERENCE_CAP`, so no cap changes; only the `context` key is new.

Nothing tied the context table to `PERSON_SECTIONS`, which is how three sections reached `main` without an entry. A structural test now walks every person section, and the "Adding a New Section" checklist in `CONTRIBUTING.md` names `link_metadata.py`.

Verified the three new tests fail with the fix stashed. `pytest` passes; `ruff`, `ruff format`, `ty` and `pre-commit` clean.

## Left out on purpose

`saved_jobs`, `jobs` and `feed` also reach `derive_context` without a table entry. Unlike the four here they have no static sibling that settles what the value should be, and `posts`, `search_results`, `main_profile` and `about` show the table is deliberately not exhaustive — those get dynamic contexts by design. I'll open a separate issue for those three rather than fold a design question into this fix.

## Synthetic prompt

> `_SECTION_CONTEXTS` in `scraping/link_metadata.py` is missing `certifications`, `skills`, `projects` and `employees`, so their references come back with no `context` key and lose dedupe ties via `_reference_score`. Add them to `_SECTION_CONTEXTS` and `_REFERENCE_CAPS` (caps at the existing default so nothing changes), add a structural test that walks `PERSON_SECTIONS` so this cannot drift again, and add the file to the "Adding a New Section" checklist in `CONTRIBUTING.md`. Leave `saved_jobs`, `jobs` and `feed` alone.

Generated with Claude Opus 5


ConnorMoss02:
The one failing check is `test_loads_full_page_on_a_slow_connection` with `assert 5 == 25`, which is #812 — the same test, same assertion, and #813 is open to widen the margin.

Not related to this diff: it touches `link_metadata.py`, `CONTRIBUTING.md` and `tests/test_link_metadata.py` only, and #831 passed the same job on the same base. The file skips locally here without a browser, so I could not rerun it, and I do not have rerun rights on the workflow.

ConnorMoss02:
Rebased onto `main` to pick up #813, so the `test` failure above was the #812 flake on the old base and is no longer in the tree. No change to the diff.

ConnorMoss02:
The rebase cleared the #812 flake, and this run hit a different one: `test_commit_authorization_expires_while_parent_pipe_stays_open`. Filed as #845 with the thread-ordering diagnosis.

Still unrelated to this diff, which touches `scraping/link_metadata.py`, `CONTRIBUTING.md` and `tests/test_link_metadata.py`. Happy to rebase again whenever you want a green run.

stickerdaniel:
Verified this end to end against live LinkedIn on the exact current base and updated head.

- `skills`, `certifications` and `projects`: 10 references each
- `employees`: 3 references
- Before: every target reference lacked `context`
- After: every reference carried exactly its source section
- Kind, URL, text, value, count and order were otherwise identical; no section errors

I rebased the branch onto current `main` and added one maintainer polish commit so the contributor checklist permits dynamic context handling rather than requiring every future section to enter the static table. The full suite passes with 3,053 tests and 60 platform skips; all hooks pass, the context-removal mutation fails all three new tests, and the independent final review found no issues.

PR #849, opened 2026-09-01
test(control): Allow scheduler margin

### LINKED ISSUE #848: [BUG] Silent-peer queue test has too little scheduler margin
**Setup:** CI on Ubuntu with `uv run pytest -n auto`.

## What Happened

`TestAuthorization::test_a_full_queue_of_silent_peers_leaves_the_owner_its_turn` requires the complete 15-peer queue to drain in less than half of the one-second production attachment window. The production calculation spends about 0.38 seconds in peer allowances; the remaining 0.12 seconds is an assertion about thread scheduling, socket accept/close overhead and runner load.

A required run completed the authenticated owner handoff correctly in 0.535 seconds, then failed only the `< 0.5` timing assertion. The owner still retained 0.465 seconds of its production window. Thirty local repetitions measured 0.398 to 0.410 seconds, confirming that the runner failure was scheduler overhead rather than a changed peer allowance.

The test should retain a substantial margin below the one-second production limit while allowing realistic scheduler overhead. A three-quarter-window ceiling leaves the owner at least one quarter of its production turn and still fails an implementation where a silent peer can consume the whole wait.

## Steps to Reproduce

1. Run `uv run pytest -n auto tests/test_process_control.py::TestAuthorization::test_a_full_queue_of_silent_peers_leaves_the_owner_its_turn` under runner load.
2. Observe that the authenticated child receives the authorization record.
3. Observe an occasional failure when elapsed time is slightly above 0.5 seconds but well below the one-second production deadline.

## Logs

```shell
AssertionError: a full queue of silent peers spent the production wait
assert 0.5347016709999934 < (1.0 / 2)
```

stickerdaniel:
Closes #848

The full 15-peer queue spends about three eighths of the one-second production attachment window in deliberate peer allowances. The test required the entire drain below one half, leaving only one eighth for thread scheduling and socket cleanup; a loaded runner completed a correct authenticated handoff in 0.535 seconds and failed the timing assertion.

Keep one quarter of the production window for the owner. Thirty local runs measured 0.398 to 0.410 seconds, the control-channel suite passes, and mutating one peer to consume the whole wait still fails the test.

## Synthetic prompt

> Make the full silent-peer queue timing test tolerate realistic scheduler overhead while preserving a substantial owner attachment window and still rejecting the original full-wait defect.

### DIFF
--- a/tests/test_process_control.py
+++ b/tests/test_process_control.py
@@ -161,9 +161,13 @@ class TestAuthorization:
             listener.send(_RECORD)
 
             assert child.readline() == _RECORD
-            # And with room to spare rather than by a hair, so a loaded machine
-            # reaches the same verdict.
-            assert elapsed < daemon_election._PREPARED_READ_SECONDS / 2, (
+            # Keep a quarter of the production wait for the owner. The peer
+            # allowances consume about three eighths; requiring the complete
+            # drain below one half left only one eighth for thread scheduling
+            # and socket cleanup, and a loaded runner spent 0.535s on a correct
+            # handoff. Three quarters still fails any peer that can spend the
+            # whole wait while leaving realistic scheduler margin.
+            assert elapsed < daemon_election._PREPARED_READ_SECONDS * 3 / 4, (
                 "a full queue of silent peers spent the production wait"
             )
         finally:


stickerdaniel:
@greptileai review

PR #855, opened 2026-09-02
test(daemon): Prove blocked control timeout

### LINKED ISSUE #845: [BUG] Flaky test: test_commit_authorization_expires_while_parent_pipe_stays_open
### Setup

CI, `uv run pytest -n auto`. Seen on [run 33456471423](https://github.com/stickerdaniel/linkedin-mcp-server/actions/runs/33456471423/job/99697771768), on a PR branch whose diff touches only `scraping/link_metadata.py`, `CONTRIBUTING.md` and `tests/test_link_metadata.py`.

### What Happened

`tests/test_daemon_election.py::TestPublishingLast::test_commit_authorization_expires_while_parent_pipe_stays_open` failed:

```
AssertionError: assert ['probe', 'prepare', 'prepared', 'logged', 'retry', 'control']
                    == ['probe', 'prepare', 'prepared', 'control', 'logged', 'retry']
```

The same test passes locally on `main` and passed on the previous run of the same branch, so this is timing rather than a regression.

`"control"` is appended by `_SuspendedParent.readline` (line 6359), which runs on the reader thread and then blocks on `release.wait()`. `"logged"` and `"retry"` are appended by the main flow. Nothing orders those two threads against each other, so the assertion pins one particular interleaving. Under `-n auto` the reader thread can be scheduled after the warning is logged, and the list comes out in the other order.

Both orderings look equally valid for what the test is actually checking, which is that commit authorization expired while the pipe stayed open.

### Steps to Reproduce

Not reliably reproducible locally; it needs the scheduling pressure of parallel workers. `uv run pytest tests/test_daemon_election.py -k commit_authorization_expires` passes on `main`.

### Logs

```
==== 1 failed, 3040 passed, 57 skipped, 1437 warnings in 143.20s (0:02:23) =====
```

ConnorMoss02:
Closes #845

`_read_control_until` starts a reader thread to call `control.readline()`, and this test's parent never returns from it, so the main flow reaches its own deadline instead of receiving a result. `"control"` is appended on that reader thread and `"logged"` on the main flow, and nothing orders the two, so the exact interleaving the assertion pinned was never guaranteed. Under `-n auto` the warning can land first, which is what CI saw.

The sibling test above it is unaffected: its parent returns from `readline` immediately, so the future resolves before the warning and the order is deterministic there.

## Changes

- `tests/test_daemon_election.py`: assert that `"control"` was recorded, and that the main flow's own sequence is exact. Its position relative to the main flow is no longer asserted.

## What the test still catches

Checked each case against the new assertion:

| Order | Result |
|---|---|
| `control` before `logged` (the previously documented order) | passes |
| `control` last (the CI failure) | passes |
| `control` between `logged` and `retry` | passes |
| `control` never recorded | fails |
| `logged` and `retry` swapped | fails |
| `retry` missing | fails |
| `prepare` missing | fails |
| `commit` leaked in | fails |

So only the cross-thread ordering is given up.

## Validation

```bash
uv run pytest tests/test_daemon_election.py
uv run ruff check tests/test_daemon_election.py
uv run ruff format --check tests/test_daemon_election.py
```

184 passed, 1 skipped (no patchright browser cache locally). Ruff clean, file already formatted.

## Synthetic prompt

> `test_commit_authorization_expires_while_parent_pipe_stays_open` in `tests/test_daemon_election.py` is flaky under `-n auto` because it asserts an exact order containing `"control"`, which is appended on the reader thread `_read_control_until` starts while the main flow races its own deadline. Relax the assertion to check that `"control"` was recorded and that the main flow's sequence is exact, keeping every other guarantee.

Generated with Claude Opus 5

### DIFF
--- a/tests/test_daemon_election.py
+++ b/tests/test_daemon_election.py
@@ -6353,7 +6353,10 @@ class TestPublishingLast:
     ):
         release = threading.Event()
         control_read = threading.Event()
+        finished = threading.Event()
         order: list[str] = []
+        outcomes: list[int] = []
+        errors: list[BaseException] = []
 
         class _SuspendedParent:
             def readline(self) -> str:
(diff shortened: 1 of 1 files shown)


ConnorMoss02:
Right, and that was a real loss. The old exact-order assertion ruled out a second control read; `in` does not, and only the position was meant to be relaxed. Changed to `order.count("control") == 1`.

The three thread interleavings still pass, a duplicate control read now fails, and a missing event or a swapped logged/retry fails as before.

184 passed, ruff clean.

PR #871, opened 2026-09-03
fix(messaging): Verify recipient before sending

### LINKED ISSUE #861: [BUG] send_message can accept an unrelated recipient
## Setup

`origin/main` at `95c3e18a`, Patchright Chromium, synthetic browser DOM reproduction. The related read-only live probe used the authenticated macOS profile with `uv run` and `confirm_send=false`.

## What Happened

`send_message` can accept the wrong compose recipient. A supplied `profile_urn` determines the compose URL, but the later recipient check only looks for the requested username or display name anywhere in visible `<main>` content. A matching name in an inbox row, sidebar, or unrelated profile card therefore satisfies the check even when the actual composer targets someone else.

The profile-page fallback has the same trust-boundary problem: `_resolve_message_compose_href()` selects the first visible `/messaging/compose/` link anywhere under `<main>`, so an unrelated card can determine the recipient before the compose page opens.

The recipient identity should be derived from the loaded target profile and remain consistent with any supplied URN, selected compose link, resulting compose URL, and composer-local recipient. Any mismatch or ambiguity must stop before editor focus, message entry, or submission.

## Steps to Reproduce

1. Render a synthetic compose page whose composer targets one person.
2. Add the requested person's name only to an unrelated visible sidebar or inbox row under `<main>`.
3. Call `_compose_page_matches_recipient()` with the requested name.
4. Observe that it returns `True` because the implementation searches the whole `<main>` and accepts substring matches.
5. On a synthetic profile page, place an unrelated visible compose link before the target profile action link.
6. Call `_resolve_message_compose_href()` and observe that the unrelated link is returned.

A read-only live probe against the approved test contact reached LinkedIn's Premium InMail modal rather than a hydrated composer. It confirmed that no editor was focused, no message text was entered, and no message was sent. It could not supply a real composer-recipient DOM sample in that account state.

## Logs

```shell
sidebar_candidate_matched=True
```

stickerdaniel:
Closes #861

`send_message` opened a compose surface and typed the caller's text into whatever editor it found. Nothing tied that editor to the person the caller named, so a stale composer, a redirect, or a recipient chip for another member could deliver the message to someone the caller never addressed, and the result still reported success.

The profile page, its compose link, the messaging URL, and the composer's own recipient identity now have to name one person before anything is focused, typed, or submitted. On any disagreement the tool returns with the editor untouched. Draft content is excluded from that identity read, so text the caller wrote cannot authorize its own delivery. The walk collects every identity up to the enclosing dialog, so a profile merely mentioned in a rendered thread refuses later sends to that person: extra identities can only tighten the check, never redirect a message.

A synthetic Chromium page drives the real production path; with the identity check mutated away, that page delivers the message to a foreign recipient, which is Issue #861 reproduced in a DOM.

## Synthetic prompt

> `send_message` types into whatever composer it finds, so the message can reach the wrong person. Bind the profile page, `profile_urn`, the compose link, the messaging URL, and the composer's own recipient identity to one person, and abort before focus, typing, or submission on any disagreement. Draft content must never authorize a recipient. Cover it with DOM tests on the real path and verify each one by mutation.

Generated with Claude Opus 5 and Claude Fable 5.1

### DIFF
--- a/linkedin_mcp_server/scraping/extractor.py
+++ b/linkedin_mcp_server/scraping/extractor.py
@@ -532,12 +532,17 @@ _MESSAGE_COMPOSER_INSPECT_JS = r"""
                 url.username ||
                 url.password ||
                 (url.port && url.port !== '443') ||
-                url.hash ||
-                !/^\/in\/[^/?#]+\/$/.test(url.pathname)
+                url.hash
             ) {
                 return null;
             }
-            return url.pathname;
+            // Everything under /in/<slug>/ belongs to that member, so the
+            // canonical path is the identity. Measured on a live profile:
+            // four of its own anchors point at /overlay/... and
+            // /recent-activity/, and reading those as an unknown member let
+            // the recipient contradict themselves.
+            const match = /^\/in\/([^/?#]+)(?:\/.*)?$/.exec(url.pathname);
+            return match ? `/in/${match[1]}/` : null;
         } catch {
             return null;
         }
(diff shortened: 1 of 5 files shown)


stickerdaniel:
Superseded by #896, which incorporates this branch's recipient verification and the final owner, route, and send-evidence contract. #896 was merged as 70e50ada.

PR #892, opened 2026-09-05
fix(references): Add contexts for jobs, saved_jobs and feed

### LINKED ISSUE #844: [BUG] saved_jobs, jobs and feed references have no context
### Setup

`main` at `991799c`. Not environment-specific; reproduces from a direct call to `build_references`.

### What Happened

Follow-up to #830, which covered `certifications`, `skills`, `projects` and `employees`. Three section names still reach `derive_context` without a `_SECTION_CONTEXTS` entry and fall through to `return heading if heading in _CONTEXT_LABELS else None`:

- `saved_jobs`, passed at `scraping/extractor.py:4203`
- `jobs`, a `COMPANY_SECTIONS` entry (`scraping/fields.py:25`)
- `feed`, passed at `scraping/extractor.py:757`

I left these out of #832 on purpose. Unlike the four there, they have no static sibling that settles what the value should be, and the table is deliberately not exhaustive: `posts`, `search_results`, `main_profile` and `about` all get dynamic contexts from `derive_context` by design.

Which direction do you want for these three? Static entries like `inbox` and `conversation`, dynamic handling like `posts`, or leave them as they are.

Either way the dedupe effect from #830 applies: `_reference_score` scores a context-less reference strictly lower, so these sections lose duplicate-url ties during cross-page dedupe.

Happy to send the PR once you have picked a direction.

### Steps to Reproduce

```python
from linkedin_mcp_server.scraping.link_metadata import build_references

raw = [{"href": "https://www.linkedin.com/jobs/view/123/", "text": "Senior Engineer"}]

for section in ("saved_jobs", "jobs", "feed", "inbox"):
    print(section, build_references(raw, section)[0])
```

`inbox` carries `'context': 'inbox'`. The other three carry no `context` key.

### Logs

n/a, not a runtime failure.

ConnorMoss02:
Closes #844

`saved_jobs`, `jobs` and `feed` reach `derive_context` with no `_SECTION_CONTEXTS` entry, so their references arrive with no `context` key. `_reference_score` ranks a context-less reference below any duplicate that has one, so these sections lose cross-page dedupe ties.

Static rather than dynamic: post permalinks are dropped before `derive_context` runs, so what survives in `feed` is people, companies and jobs, and the `posts` labels would be wrong for two of the three.

- `jobs` → `"jobs"`, matching `employees`
- `saved_jobs` → `"saved jobs"`, matching `job_posting`
- `feed` → `"feed"`

The existing guard only walked `PERSON_SECTIONS`, which is why it stayed green through all three. It now walks every declared section. Names live in `PERSON_SECTIONS`, `COMPANY_SECTIONS` and `_REFERENCE_CAPS`, and `saved_jobs` is in none of them, so the test names it explicitly.

One unrelated failure, `test_daemon_descriptor.py::TestEndpointSpelling`, reproduces on clean `main` at 13734ba.

### DIFF
--- a/linkedin_mcp_server/scraping/link_metadata.py
+++ b/linkedin_mcp_server/scraping/link_metadata.py
@@ -87,6 +87,9 @@ _SECTION_CONTEXTS = {
     "job_posting": "job posting",
     "inbox": "inbox",
     "conversation": "conversation",
+    "jobs": "jobs",
+    "saved_jobs": "saved jobs",
+    "feed": "feed",
 }
 
 _DEFAULT_REFERENCE_CAP = 12
--- a/tests/test_link_metadata.py
+++ b/tests/test_link_metadata.py
@@ -592,6 +592,27 @@ class TestBuildReferences:
             }
         ]
 
+    def test_names_the_context_for_jobs_saved_jobs_and_feed(self):
+        """The structural guard below only proves a context key exists, so it
+        survives a wrong label. These are the values themselves."""
+        raw: list[RawReference] = [
+            {
+                "href": "https://www.linkedin.com/jobs/view/123/",
+                "text": "Senior Engineer",
+            }
+        ]
+
+        contexts = {
+            section: build_references(raw, section)[0]["context"]
+            for section in ("jobs", "saved_jobs", "feed")
+        }
+
+        assert contexts == {
+            "jobs": "jobs",
+            "saved_jobs": "saved jobs",
+            "feed": "feed",
+        }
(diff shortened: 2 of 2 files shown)


ConnorMoss02:
Added `test_names_the_context_for_jobs_saved_jobs_and_feed` asserting the three values. Changing `"saved jobs"` to `"wrong label"` fails it while the structural guard stays green.

Kept both: the guard catches a section arriving with no context, the new test protects the labels.

Issue #891, opened 2026-09-05
Owner loss hides a possibly delivered message


stickerdaniel:
### Setup

Default stdio startup, which uses the frontend/owner split. Present on `main`.

### What Happened

When the shared browser owner disappears mid-call, `ProxyRecoveryMiddleware.on_call_tool` (`daemon_proxy.py`) elects a replacement and then decides whether to repeat the call. For a mutating tool whose request had already left this process, it refuses to repeat and re-raises, and the comment there is right about why: nothing in the protocol says whether the departed owner had already sent the message, and guessing wrong sends it twice.

The refusal is correct. What reaches the caller is not. `OwnerUnreachableError` is a plain `Exception` and the server runs with `mask_error_details=True`, so the client sees a masked generic tool error. It carries no `status`, no `retry_safe`, and no hint that a message may already have been delivered. An autonomous client reads a masked error as transient and retries, which is exactly the second delivery the refusal was protecting.

This is not the deadline case in #889: there the tool's own scope is cancelled and no result can survive. Here the frontend process is alive and answering; it just answers with nothing the caller can act on.

Concrete path:

1. `send_message` clicks Send and LinkedIn accepts the message.
2. The owner process dies before the result comes back.
3. `failure.nothing_was_sent` is false, because the request had been dispatched.
4. `a_repeat_could_change_something` is true from the tool's `destructiveHint` annotation.
5. The middleware elects a replacement owner and re-raises.
6. The client sees a masked error and retries.

`connect_with_person` has the same shape.

### What a fix has to decide

The small half is available now and is the repository's existing pattern: raise a `ToolError` whose text names the risk, so `mask_error_details` does not reduce it, and the caller is told the call may have taken effect. `tools/person.py` and `tools/post.py` already do this for the same reason.

The larger half is a design decision and is why this is an issue. Synthesizing a real result (`status: "send_unconfirmed"`, `retry_safe: false`) would let a caller branch on the same fields it uses everywhere else, but the middleware sits below the tools and would have to know each mutating tool's result schema, including a `url` the departed owner never reported. A shared "dispatched, outcome unknown" shape across every destructive tool would solve that, and inventing one is not a patch to the proxy.

### Steps to Reproduce

Start the server, call `send_message` with `confirm_send=true`, and kill the owner process after the request is dispatched but before it returns. The client receives a masked tool error rather than an outcome.

PR #895, opened 2026-09-07
fix(daemon): Surface owner-loss risk as ToolError

Thespaceblade:
When a mutating call may already have run and the shared browser owner is unreachable, raise ToolError so mask_error_details keeps the warning instead of a masked generic error that invites a blind retry (#891).

### DIFF
--- a/linkedin_mcp_server/daemon_proxy.py
+++ b/linkedin_mcp_server/daemon_proxy.py
@@ -45,6 +45,7 @@ from typing import TYPE_CHECKING, Any
 
 import httpx
 import mcp.types as mt
+from fastmcp.exceptions import ToolError
 from fastmcp.server.middleware import CallNext, Middleware, MiddlewareContext
 from fastmcp.tools import ToolResult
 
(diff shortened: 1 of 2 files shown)


stickerdaniel:
Thanks for this. The `ToolError` observation is right, and checking it is what made me read the masking path properly: `ToolError` subclasses `FastMCPError`, which `server.py` re-raises untouched, so the text does survive `mask_error_details`.

I shipped #1008 for #891 instead, so I'm closing this one. Two things it does that this PR doesn't:

It covers both raises. This changes the branch where a replacement was found, but the earlier `if await self._backend.recover(...) is None: raise` still re-raises a plain `OwnerUnreachableError`, which is still masked. That is the same double-send on the path where the election finds nobody.

And it answers with structure rather than prose: `status: "outcome_unknown"` and `retry_safe: false`, so a client doesn't have to parse an English sentence to learn that a retry can deliver the message twice.

An external review of #1008 turned up a third path that neither approach covers, where cleanup replaces the tagged exception and the middleware never recognises the failure at all. That is #1010 if you want a look.

PR #931, opened 2026-09-13
fix(config): stop config loading from parsing the test runner argv

### LINKED ISSUE #929: [BUG] Test suite fails under any non-bare pytest invocation (config loader parses live sys.argv)
### Setup

macOS, running the test suite locally with `uv run pytest` (per CLAUDE.md/CONTRIBUTING.md dev commands). Not MCP-client-specific — this is a test-suite/config-loading bug, not a runtime bug against a real client.

### What Happened

73 tests fail whenever `pytest` is invoked with any extra CLI argument — including the project's own documented `uv run pytest --cov` coverage command, or common flags like `-k`/`-q`/`-v`. All 73 failures share one root cause and are not independent flakes:

`load_from_args()` in `linkedin_mcp_server/config/loaders.py:758` calls:

```python
args = parser.parse_args()
```

with no explicit `argv`, so it parses the **real live `sys.argv`** — i.e. pytest's own process arguments — instead of an empty/injected argument list. `argparse` has no tolerance for unrecognized arguments, so any token pytest received that isn't one of this project's own CLI flags (a marker expression, `-q`, `--cov`, an IDE test-runner flag, a CI wrapper flag like `--junitxml`) makes the parser call `sys.exit(2)`.

That `SystemExit` unwinds through `get_config() → _installer_temporary_parent() → _create_installer_temporary_root() → _run_in_daemon_thread()` and fails every test that reaches config loading for the first time in that process (config is cached in a module-level singleton, so whichever test happens to trigger the first fresh `load_config()` call in a given run pays for it). In practice that's most of `TestPatchrightInstallStreaming` and `TestCredentialRedaction` in `tests/test_bootstrap.py`, plus a handful of related classes — 73 tests total in the runs I reproduced this with.

Confirmed reproducible on stock `main` (74a1bf3) — `config/loaders.py` is untouched on the branch I was working from, so this isn't specific to any local changes.

### Steps to Reproduce

```bash
git clone https://github.com/stickerdaniel/linkedin-mcp-server.git
cd linkedin-mcp-server
uv sync --group dev
uv run pytest -q   # or: uv run pytest --cov, or -k "<anything>"
```

Any of these trigger the failures. A completely bare `uv run pytest` (zero extra tokens) does **not** trigger it, because `sys.argv[1:]` is then empty and `parser.parse_args()` has nothing unrecognized to reject — which is why this can look like it "usually passes" depending on how someone invokes it.

Minimal isolated repro of the underlying collision:

```bash
uv run pytest -k "test_mirror_credentials_are_not_logged"
```

### Logs

```
tests/test_bootstrap.py:5366: in test_mirror_credentials_are_not_logged
    await bootstrap._run_patchright_install("--no-shell", line_callback=seen.append)
linkedin_mcp_server/bootstrap.py:2326: in _run_patchright_install
    await _install_under_supervision(...)
linkedin_mcp_server/bootstrap.py:2353: in _install_under_supervision
    temporary = await _run_in_daemon_thread(...)
linkedin_mcp_server/bootstrap.py:2065: in _create_installer_temporary_root
    parent = _installer_temporary_parent()
linkedin_mcp_server/bootstrap.py:2010: in _installer_temporary_parent
    configured_parent = get_config().browser.installer_temp_dir
linkedin_mcp_server/config/__init__.py:23: in get_config
    _config = load_config()
linkedin_mcp_server/config/loaders.py:893: in load_config
    config = load_from_args(config)
linkedin_mcp_server/config/loaders.py:758: in load_from_args
    args = parser.parse_args()

self = ArgumentParser(prog='pytest', ...)
message = 'pytest: error: unrecognized arguments: -k test_mirror_credentials_are_not_logged\n'
E   SystemExit: 2

Full suite via `uv run pytest -q`:
==== 73 failed, 3368 passed, 58 skipped, 1613 warnings in 426.55s (0:07:06) ====
```

Suggested fix direction: have `load_from_args()` accept an explicit `argv` parameter (defaulting to `None`/`sys.argv[1:]` only at the real CLI entry point), and/or have tests that reach `get_config()` inject a clean argv via `monkeypatch.setattr(sys, "argv", [...])` the way `tests/test_config.py` and a few others already do — the fix just needs to cover the config path these 73 tests go through too.

feiiiiii5:
## Summary

Fixes #929: `load_from_args()` called `parser.parse_args()` with no explicit argv, so it parsed the live `sys.argv` of whatever process loaded configuration. Under pytest that is pytest's own command line, and any extra token (`-q`, `-k`, `--cov`, CI `--junitxml`) made argparse `exit(2)` in whichever test first reached `get_config()` — 73 failures that vanish under a bare `pytest`, which is why this looked intermittent.

## Changes

- `linkedin_mcp_server/config/loaders.py`: `load_from_args(config, argv=None)` — explicit argv goes straight to `parse_args(argv)`; the `None` default preserves the exact historical behavior for the real CLI entry point.
- `tests/conftest.py`: new autouse `isolate_process_argv` fixture pinning `sys.argv` to `["linkedin-mcp-server"]`, sibling to the existing `ignore_the_developers_environment` ambient-state fixture. Tests that prove CLI parsing set `sys.argv` themselves (as `test_config.py` does throughout); a test-body assignment always wins over a fixture, so only the tests that never mention argv are affected. This is safe for the currently-green suite by construction: any extra argv token is fatal today, so no passing test can depend on one.
- `tests/test_config.py` (`TestLoaders`): three regression tests — explicit argv wins over a dirty runner argv (defaults intact + flag applied), explicit argv stays strict (unknown flag still `SystemExit`), and the bare-call hazard is pinned as the documented reason the fixture exists.

## Testing

- Reproduced pre-fix in isolation against stock `main` (`74a1bf3`): `load_config()` with `sys.argv == ["pytest", "-q", "-k", ...]` → `SystemExit(2)` (`pytest: error: unrecognized arguments`), matching the issue's traceback shape.
- Post-fix in the same harness: explicit clean/flag argv loads and applies under a dirty runner argv; unknown flags still exit; bare call still parses own argv (prod behavior unchanged); full `load_config()` succeeds once argv is clean (i.e. what the fixture guarantees in-suite).
- New tests are hermetic (explicit `monkeypatch.setattr("sys.argv", ...)` in each); the 2 network-style tests in this area are untouched. Full-suite confirmation (`uv run pytest -q`, the issue's own repro) left to CI — it needs the project's dev group and browser fixtures.

Signed-off-by: fei <204683769+feiiiiii5@users.noreply.github.com>


stickerdaniel:
Thanks for the contribution. This cleaned up pytest's argv, but left embedded `get_config()` calls exposed to host arguments. I fixed the boundary in #942. Closing as superseded.

PR #963, opened 2026-09-13
docs: Run the test suite once per session

rimonhanna:
Adds a `## Tests` rule documenting that the suite should run once per session rather than after every change, since parts of it exercise a real LinkedIn session against the profile under `~/.linkedin-mcp/` and spend the account's rate budget. Mirrors a rule already adopted downstream after a session that ran the suite after each of a dozen commits got the account throttled.

## Synthetic prompt

> Add a rule to the Tests section of AGENTS.md: run the full test suite once per session, at the end, not after every change, because parts of it drive a real LinkedIn session against the profile under `~/.linkedin-mcp/` and spend the account's rate budget. Document a mid-flight alternative (ruff check, ruff format --check, ty check, the two --check scripts under scripts/, and at most the single pinned test module) and note subagents follow the same rule.

Generated with Claude Opus 5


rimonhanna:
@stickerdaniel small docs-only PR, independent of the stack.

stickerdaniel:
Thanks @rimonhanna, but I'm closing this because the suite never touches the real profile or LinkedIn: since #650 the autouse `isolate_profile_dir` fixture in `tests/conftest.py` points every test at a profile under `tmp_path`.

Issue #838, opened 2026-08-31
[BUG] CPython 3.12.4 can crash during concurrent daemon election


stickerdaniel:
**Setup:** shared daemon mode on Windows using uv's managed CPython 3.12.4, under eight simultaneous frontend processes.

## What Happened

`TestRealOwner::test_many_clients_starting_at_once_elect_exactly_one_owner` intermittently terminates one frontend with Windows status `0xC0000005` (`STATUS_ACCESS_VIOLATION`). Two CI runs on the same commit produced one pass and one crash; six of seven observed 3.12.4 jobs passed. Current 3.12, 3.13, and 3.14 passed the same test.

`faulthandler` placed daemon reader threads in `daemon_election.py`, but the captured output was truncated before the native fault location. No repository stream close/read defect was found. The closest upstream match is python/cpython#110052 and #124878, a Windows access-violation race involving daemon threads and file I/O, but the surviving main-thread state does not prove this is the same finalization defect. A python.org build comparison or native minidump is still needed.

The exact Windows CPython 3.12.4 job now skips only this stress test. Every other declared-floor test still runs there, and current 3.12 through 3.14 continue to exercise the eight-client election.

## Steps to Reproduce

1. Install uv's managed CPython 3.12.4 on Windows.
2. Run the following test repeatedly or in parallel:

   `uv run pytest tests/test_daemon_election.py::TestRealOwner::test_many_clients_starting_at_once_elect_exactly_one_owner`

3. Observe that a low-probability run exits natively instead of producing a Python exception.

## Logs

```shell
frontend.returncode = 3221225477  # 0xC0000005

Thread ...
  File "linkedin_mcp_server/daemon_election.py", line 706 in collect

Thread ...
  File "threading.py", line 997 in start
  File "linkedin_mcp_server/daemon_election.py", line 1721 in _read_owner_verdict
  File "linkedin_mcp_server/daemon_election.py", line 1740 in _await_prepared
  File "linkedin_mcp_server/daemon_election.py", line 974 in _spawn
  File "linkedin_mcp_server/daemon_election.py", line 362 in obtain_owner
```

framegrabber:
Not specific to 3.12.4. The same crash — same `0xC0000005`, same daemon reader threads parked in `daemon_election.py:706` — has now failed three main-branch runs in two days on the matrix's **3.12.10**:

- https://github.com/stickerdaniel/linkedin-mcp-server/actions/runs/34930399721
[...]

PR #1011, opened 2026-09-18
ci(daemon): Capture a native stack for #838

stickerdaniel:
## Problem

#838 could not be diagnosed because nothing in CI could describe the fault.

`faulthandler` cannot: with `_both_ends` (#996) finally keeping the head of the report, a soak caught the access violation and printed a *complete* dump — banner, five threads, main thread last — with not one marked `Current thread`. `faulthandler_dump_traceback` takes the faulting thread from `PyGILState_GetThisThreadState()` and `_Py_DumpTracebackThreads` marks a thread only on pointer equality with it, so an unmarked dump means that call returned `NULL`. The thread that faults holds no Python thread state.

And no dump existed either. Armed as it was, one soak caught the crash and uploaded nothing. The runner image leaves Windows Error Reporting disabled, and a disabled WER never consults `LocalDumps`.

## Solution

Enable WER before arming it, and register `procdump` under `AeDebug` too — the process that dies is a grandchild, so nothing can attach to it in advance. Print the dump folder and the system's fault record, because the upload step reports "no files were found" identically for a clean run and a silent one. Clear `SEM_NOGPFAULTERRORBOX` in the test frontend: while it is set, `UnhandledExceptionFilter` reaches none of those mechanisms and the process just exits carrying the exception code.

## Verification

That last bit was the one that mattered; the error mode measured 0x0003. The next run produced full dumps and an event-log entry naming `_wmi.pyd`, which is the root cause, fixed in #1013.

The procedure is written up as a runbook, `docs/windows-crash-dumps.md`:
the three preconditions that each yield an empty artifact, and how to
symbolise a dump on macOS.

Refs #838.

## Synthetic prompt

> #838's soak catches the Windows access violation but no dump is ever written and faulthandler cannot name the faulting thread. Make `election-soak.yml` actually produce a native stack: enable Windows Error Reporting before arming LocalDumps, register procdump as the postmortem debugger for the grandchild that dies, print the dump folder so an empty upload differs from a clean run, and clear SEM_NOGPFAULTERRORBOX in the test frontend so UnhandledExceptionFilter reaches any of it.

Generated with Claude Opus 5


stickerdaniel:
Checked the masked-failure finding. The loop does not continue past a failing round.

GitHub Actions runs a `shell: bash` step as `bash --noprofile --norc -eo pipefail {0}`. Under `-e`, `uv run pytest` in a plain `for` body aborts the shell immediately: the command is not in a condition, not negated, and not followed by `&&` or `||`, so none of the exemptions apply. Reproduced:

```
$ bash --noprofile --norc -eo pipefail -c 'for i in 1 2 3; do echo "::group::round $i"; false; echo "SHOULD NOT PRINT"; done'
::group::round 1
$ echo $?
1
```

The `::endgroup::` line never runs and the step fails, so `if: failure()` fires.

The observed runs agree. Runs 35412200013, 35412205154 and 35412210699 each ended in `failure` with one artifact, and those artifacts are what produced the dumps this investigation was built on. The upload path is not hypothetical; it is the path the root cause came through.

On the ProcDump download: no checksum, correct. It is an HTTPS fetch from `download.sysinternals.com` onto an ephemeral runner, in a workflow that only runs on manual dispatch and ships nothing. Pinning a hash would mean pinning a binary Microsoft republishes under the same URL, which breaks the workflow on every refresh without making the runner safer.

stickerdaniel:
Applying the same knowledge-policy correction here, which nothing flagged but the policy reads the same way.

Four comment blocks in this PR carried measurements and external references: run ids, the measured `0x0003` error mode, and CPython `faulthandler` internals. `docs/knowledge-policy.md` puts those in documents with their source and date, and keeps the local comment short.

Capturing and symbolising a native dump is an operational procedure, so it is now a runbook: `docs/windows-crash-dumps.md`, with prerequisites, commands and completion criteria. The three preconditions are worth having written down, because each one alone yields an empty artifact and they look identical in the log: the error-mode bit, WER being off in the runner image, and the faulting process being a grandchild nothing can attach to in advance.

The four comments are now short warnings pointing at it.

stickerdaniel:
Added a signature check for the downloaded `procdump`, and left the loop finding alone.

On the download: the concern is fair in substance, because this binary is registered as the debugger Windows starts on any unhandled exception. A pinned hash is the wrong instrument though, since Sysinternals republishes the same URL and the pin would fail on every refresh while proving nothing the signature does not. The step now checks Authenticode: status `Valid`, and the signer subject containing `O=Microsoft Corporation`.

Verified in both directions on `windows-latest`:

- run 35426602637, three rounds, green. `procdump signed by CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US`.
- the same workflow with the expected signer changed to a name Microsoft does not use: the step fails and the soak never starts. Without that, the check would read as protection while asserting something every download satisfies.

On the masked soak failures: still not reproducible. GitHub runs `shell: bash` as `bash --noprofile --norc -eo pipefail`, so a failing `uv run pytest` in a plain `for` body ends the shell. The reasoning is in my earlier comment, and the run above is a further data point: when a step in it did fail, `Upload the crash dump` ran.

stickerdaniel:
An independent review found a scope problem here. Fixed in `c24647d1`.

**Clearing `SEM_NOGPFAULTERRORBOX` was not confined to the soak.** `_INSPECT_OWNER` is the program `_run_frontend` executes, and nineteen tests go through that helper. The Windows CI job runs several of them with LocalDumps armed but without `DontShowUI`, so I was turning crash reporting back on in a job that has not armed the setting which keeps a fault from waiting for a click. It is now gated on `LINKEDIN_MCP_ELECTION_SOAK`, which only the soak workflow sets. Verified that the guard evaluates false without the variable and true with it, and that the soak still passes end to end.

**Two runbook claims were wrong.** Microsoft documents that LocalDumps is skipped while an automatic debugger is configured, so ProcDump is the capture route, not one of two, and the `DumpCount`/`DumpType` settings do not describe what lands — `-ma` writes full-memory dumps, which are large and should be treated as potentially sensitive. The `Arm a crash dump` step earns its place through `DontShowUI`, not redundancy.

I also over-read run 35367449209. It caught the violation with LocalDumps armed and uploaded nothing, but the error mode was suppressing reporting on that run too, so it does not isolate `Disabled` as the cause. Recorded as an observation now, not a mechanism.

**On signatures.** My earlier comment claimed a hash proves nothing more than the signature. That is wrong: a digest rejects changed bytes, while the signature accepts whatever Microsoft signs next. The trade is still the right one here, because Sysinternals republishes the same URL, but the step now prints the accepted version and SHA-256 so a capture stays attributable. Last run: ProcDump 12.01, `D1FC99AE304BD1D2BF28ABEB62531DA959E2431916194981B88C958FD713A8E6`.

The masked-soak-failure finding was independently checked and confirmed not to reproduce.

stickerdaniel:
Final signer-validation note: the first exact-OID version still skipped multi-valued RDNs. That allowed a hidden second organization in `CN=Tool+O=Contoso Ltd` next to a standalone Microsoft organization.

The merged check fails closed on every multi-valued RDN, then requires exactly one single-valued Organization RDN (`2.5.4.10`) whose value ordinal-equals `Microsoft Corporation`. Its runtime self-test covers the CN lookalike and the multi-valued hidden-O case. The real Microsoft signer uses separate single-valued RDNs and passed the final `windows-latest` soak. The targeted final reviewer reported no remaining finding.

PR #1013, opened 2026-09-19
fix(session): Name the Windows arch without WMI

### LINKED ISSUE #838: [BUG] CPython 3.12.4 can crash during concurrent daemon election

stickerdaniel:
## Problem

#838: a frontend dies with `0xC0000005` mid-election on Windows CPython 3.12. Six of seven 80-round soaks hit it.

`platform.system()` and `platform.machine()` are both `platform.uname()`, and on Windows `uname()` has no `os.uname()` to read, so it fills every blank itself with `win32_ver()` and `_get_machine_win32()` — a WMI query each.

CPython 3.12's `_wmi.exec_query` hands a `CreateThread` worker a pointer to a struct on its own stack, waits 1000ms for COM and 100ms for the connection, and on timeout closes the handles and returns anyway. The worker is still running and reads through the frame that just went away. Captured: `EXCEPTION_ACCESS_VIOLATION_READ` at `mov rcx, qword [rsi + 0x8]` in `_wmi.pyd`, `rsi` pointing into a thread stack.

That worker has no Python thread state, which is why every `faulthandler` report of this named every thread except the one that faulted.

GH-130727 fixed it upstream by copying the struct. It is in 3.13 and 3.14 and was never backported to 3.12 — exactly the version split this issue shows. Sixteen processes initialising COM together is enough to reach those timeouts.

## Solution

`get_runtime_id()` decides Windows from `sys.platform`, not from `platform.system()`, and takes the architecture from `PROCESSOR_ARCHITEW6432` / `PROCESSOR_ARCHITECTURE`. Those are `platform`'s own fallback, and `_normalize_arch` maps their values onto what a WMI reply maps onto, so the id — which names a directory — does not move.

The first commit here got this wrong: it avoided `platform.machine()` but still called `platform.system()`, and soaking it crashed in round 2. The test now refuses `platform.uname` itself, which catches either route in.

## Verification

254 soak rounds on 3.12.10 with no access violation. Before this, six of seven 80-round runs crashed, and the first wrong fix crashed at rounds 2, 5 and 6; at that rate 254 rounds should have produced about a dozen.

Three tests, each checked against two mutations: no fix, and that first wrong fix. 273 tests across every file that touches `get_runtime_id`.

`platform.uname()` was instrumented and the frontend's whole import set run against it: zero calls, and zero from `get_runtime_id()` on the Windows branch. So nothing else in the process reaches the query, including through dependencies.

The id cannot move. `uname()` reaches `system = 'Windows'` through `if system in ('win32', 'win16')`, and leaves `machine` exactly as `_get_machine_win32()` returned it. That table yields `x86`, `ARM`, `ia64`, `AMD64`, `ARM64`, which `_normalize_arch` lowercases onto the same values the environment gives — including the WOW64 case, where `PROCESSOR_ARCHITEW6432` carries what the query would have reported.

One separate defect surfaced during the soak and is filed as #1014; it is not this crash.

The decision is recorded in `docs/decisions/2026-09-19-windows-runtime-identity.md`,
with the evidence and measurements in `docs/windows-wmi-crash.md`.

Closes #838.

## Synthetic prompt

> #838's Windows access violation comes from WMI: `platform.uname()` runs two WMI queries on Windows, and CPython 3.12's `_wmi` worker thread reads the caller's stack frame after the caller times out and returns. Make `get_runtime_id()` decide Windows from `sys.platform` and read the architecture from the PROCESSOR_ARCHITE* environment variables, keeping the normalized runtime id identical, and cover it with tests that refuse `platform.uname` outright.

Generated with Claude Opus 5


stickerdaniel:
Not acting on the docstring finding, because the policy it cites does not exist.

I grepped every Markdown file in the repository for "Decision Record", "knowledge policy" and "knowledge-placement". There are no matches. `CLAUDE.md` has no section on where rationale belongs, and neither does `AGENTS.md` or `CONTRIBUTING.md`.

The house style is the opposite of what the comment describes. About 105 functions in `linkedin_mcp_server/` carry docstrings of fourteen lines or more, and several of them do exactly what this one does: `browser_downgrade.a_version_can_be_asked_for` explains a Chromium build flag, `config/loaders._env` records measured substitution behaviour, `daemon_proxy._the_owner_answered` runs to 56 lines. `CLAUDE.md` itself names upstream issue numbers and measured values inline.

The reason to keep it next to the code is that the invariant is not self-evident. `platform.system()` looks free. Nothing at the call site suggests it runs a WMI query, and the first version of this fix avoided `platform.machine()` while still calling `platform.system()` first, which crashed in the soak at round 2. A short warning would not have prevented that. The instruction and the version split are what make the constraint checkable by the next reader.

stickerdaniel:
Correcting myself: the policy does exist and this finding is right.

`docs/knowledge-policy.md` has been on `main` since #953. My grep missed it because the checkout I ran it in predates that merge, and I asserted from that a repository-wide fact I had not actually checked against `main`. The comment above is wrong in its central claim, and the reasoning I built on it does not survive.

Reading the policy as written, the docstring violates three clauses. It is externally justified and repository-spanning, so the decision belongs in a dated record under `docs/decisions/`. It carries measurements and upstream references, which belong in a mutable document with their source and date. And the local comment is meant to be short.

Restructuring accordingly.

stickerdaniel:
An independent review found a real regression in this PR. Fixed in `001780fb`.

**The identity could move.** `platform._get_machine_win32` queries WMI *first* and reads `PROCESSOR_ARCHITEW6432`/`PROCESSOR_ARCHITECTURE` only when that fails. I had kept the fallback and dropped the query, so a process whose environment lacks those variables — a service, or an MCP host that sanitises what it passes down — went from `windows-amd64-host` to `windows-unknown-host`. That renames the directory holding its runtime profile and the session inside it stops being found. Worse, my third test asserted `windows-unknown-host` for exactly that case, so it enshrined the regression instead of catching it.

The fix adds `GetNativeSystemInfo` as the last resort. It is WMI-free, and `Win32_Processor.Architecture` and `SYSTEM_INFO.wProcessorArchitecture` are one enumeration, so the table CPython indexes with the WMI reply reads the kernel's answer unchanged. It is asked last so no process that already had an answer gets a new one.

Three mutations, each caught by exactly one test: fallback removed (the state I nearly shipped), precedence inverted, table drifted from CPython's.

**Two documentation claims were also wrong.** GH-130727 reached users in 3.13.4 on 2025-06-03, so 3.13.0 through 3.13.3 carry the defect too — a branch is not a release. And "the timeouts are only reached under contention" is too strong: the same 3.12 source describes a five-second delay when the caller lacks permission to connect. Both corrected, along with the claim that no alternative to WMI existed; `IsWow64Process2` is another.

One case is now recorded as unestablished rather than asserted: x64 under emulation on an ARM64 host, where what the WMI query used to return has not been measured.

stickerdaniel:
Final correction after the follow-up review: `GetNativeSystemInfo` was not sufficient either.

Under x86/x64 emulation on ARM64, it reports the compatibility architecture to that application. Reading it last did not fix the stripped-environment case; reading the environment first also kept the ordinary x64-emulation case on `AMD64`, while the old WMI query named the ARM64 platform. Both would move the runtime-profile directory.

The merged implementation uses `IsWow64Process2` and reads its `nativeMachine` output before either environment variable, mirroring the old WMI-first authority. `processMachine` is ignored. Old Windows without that API accepts `GetNativeSystemInfo` only for x86/AMD64, then falls back to the environment.

The final tests distinguish AMD64 process/environment from ARM64 native machine, mutation-check environment-first and process-output mistakes, and execute the real ctypes bridge on `windows-latest`. The final `sol-reviewer` pass found no remaining issue at the merged head.

PR #1019, opened 2026-09-19
test(control): Measure peer drain directly

### LINKED ISSUE #848: [BUG] Silent-peer queue test has too little scheduler margin

stickerdaniel:
## Problem

The silent-peer queue test used an aggregate wall-clock ceiling. Under runner load it measured scheduler and timeout overshoot, while its one-second drain budget also differed from production's 30-second budget.

## Solution

Observe server-side peer closes directly with portable socket wrappers. Assert that the owner attaches inside the production window, all silent peers are dropped first, and at least one consecutive drop interval stays within the expected per-peer allowance. Remove the aggregate timing threshold.

## Verification

- target test: 30/30 repetitions
- 239 process-control and election tests
- four targeted production mutations
- Ruff, format, and ty

Closes #848

## Synthetic prompt

> Replace the flaky aggregate silent-peer timing assertion with portable observations of production-budget peer drops, owner ordering, and systematic per-peer delay. Prove the test against measured regressions without changing production code.

Measured and designed with Claude Fable 5.1. Generated with GPT-5.6 Sol.


stickerdaniel:
Final implementation note: review strengthened the interval invariant after the original PR description was written. The merged test requires at least 11 of 14 consecutive peer-drop intervals to stay below 1.5 times the production allowance, not merely one interval. It also records the synchronized owner-publication boundary and checks that timestamp against the absolute one-second deadline.

The final head passed the full cross-platform CI matrix and the targeted late-publication and repeated-slowdown mutations.

